PR #121, opened 2016-08-24
Changed placeholdertext login

DennisVinke:
Changed the placeholder text for the username and password because it was confusing the members using the Omnomcom.


jonathanjuursema:
Why? :(

But specifically for the username, on the new site the default way of logging in is with e-mail. Only after linking your utwente account yourself you can use it. Taking this into account it makes more sense to show an e-mail as default username.

DennisVinke:
The Omnomcom should only be usable after a student account is linked and with that in mind it would make more sense to show a student number placeholder for the Omnomcom login screen. The correct staple battery horse as a placeholder text is genuinely confusing people and almost all people asked if they should put in their password there and if they could just use their student number instead of typing in a mailaddres..

thijsdortmann:
I agree with Dennis on the confusion caused by correct horse battery staple, as many people don't read xkcd (even though they should...). The new placeholder should start with a small s instead of a capital S though, as this is the correct format of student numbers.

jonathanjuursema:
Agreed, but I think the username placeholder should remain an e-mail address.

PR #291, opened 2016-12-07
Auto achieve

Peeterus:
- adds a scheduled command (daily) to hand out automatic achievements
- achievements can be set to automatic and can be given a user select SQL query by admins
- board can see but not edit these settings


jonathanjuursema:
Rejected, feedback via Slack.

PR #358, opened 2017-03-06
added automatic gifting for 'big kid' achievement

Peeterus:
title says it all


Peeterus:
also adds checks to achievement gallery on obtained achievements

Peeterus:
fixed achievements jumping around in the gallery on hover when using firefox

PR #665, opened 2018-03-08
Pwned Passwords Notifications

jonathanjuursema:
Adds Pwned Passwords functionality using the [Pwned Passwords API](https://haveibeenpwned.com/Passwords) (including k-anonymity feature).

* You can no longer change your password to a password that is present in a known data breach.
* If you are currently using an exposed password, you will receive an e-mail after you've logged in with that password. That e-mail will only be sent ONCE per user and exposed password combination.

There will be an [accompanying wiki page](https://wiki.proto.utwente.nl/ict/pwned-passwords) with more info and security implications.

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -578,7 +578,7 @@ class AuthController extends Controller
         }
 
         if ($user != null && Hash::check($password, $user->password)) {
-            if ((new PwnedPasswords())->setPassword($password)->isPwnedPassword() && HashMapItem::where('key', 'pwned-pass')->where('subkey', $user->id)->first() === null) {
+            if (HashMapItem::where('key', 'pwned-pass')->where('subkey', $user->id)->first() === null && (new PwnedPasswords())->setPassword($password)->isPwnedPassword()) {
                 Mail::to($user)->queue((new PwnedPasswordNotification($user))->onQueue('high'));
                 HashMapItem::create(['key' => 'pwned-pass', 'subkey' => $user->id, 'value' => date('r')]);
             }
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -9,7 +9,10 @@ use PragmaRX\Google2FA\Google2FA;
 use Adldap\Adldap;
 use Adldap\Connections\Provider;
 
+use nickurt\PwnedPasswords\PwnedPasswords;
+
 use Proto\Mail\PasswordResetEmail;
+use Proto\Mail\PwnedPasswordNotification;
 use Proto\Mail\UsernameReminderEmail;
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/AuthController.php
```diff
@@ -572,6 +578,10 @@ public static function verifyCredentials($username, $password)
         }
 
         if ($user != null && Hash::check($password, $user->password)) {
+            if ((new PwnedPasswords())->setPassword($password)->isPwnedPassword() && HashMapItem::where('key', 'pwned-pass')->where('subkey', $user->id)->first() === null) {
```
thijsdortmann: For efficiency, maybe do the HaspMapItem check first before doing the isPwnedPassword-check, as we can do the HasMapItem check locally.

app/Mail/PwnedPasswordNotification.php
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Build the message.
+     *
+     * @return $this
+     */
+    public function build()
+    {
+        return $this
+            ->subject('IMPORTANT SECURITY NOTICE: The password you use with S.A. Proto was exposed in a data breach.')
```
thijsdortmann: Maybe change the subject to something that immediately makes clear that this wasn't a data breach of Proto's data.

app/Http/Controllers/AuthController.php
```diff
@@ -397,6 +400,9 @@ public function passwordChangePost(Request $request)
             } elseif (strlen($pass_new1) < 10) {
                 $request->session()->flash('flash_message', 'Your new password should be at least 10 characters long.');
                 return view('auth.passchange');
+            } elseif ((new PwnedPasswords())->setPassword($pass_new1)->isPwnedPassword()) {
```
thijsdortmann: I haven't been able to validate this with a quick Google: does this library automatically use the k-anonimity feature? As I don't see a configuration parameter for this specified anywhere.

jonathanjuursema: [Yes it does!](https://github.com/nickurt/laravel-pwned-passwords/blob/master/src/PwnedPasswords.php#L86)

app/Http/Controllers/AuthController.php
jonathanjuursema: Agreed.

app/Mail/PwnedPasswordNotification.php
jonathanjuursema: I'm not sure if that is the right approach. It is stated very clearly in the e-mail, and perhaps this subject is a better trigger to read the e-mail.

thijsdortmann: Good point, agreed :)

PR #948, opened 2019-04-25
Daily helper notifications. Fixes #911.

### LINKED ISSUE #911: Batch helper mails
I think it would be more efficient and less spammy to group the "hey this event needs your help" emails into a daily (if any are sent that day) in to a "hey these events need your help" that is sent at midnight also spotlighting other events that still need help,

Thanks

thijsdortmann:


resources/views/emails/dailyhelpermail.blade.php
```diff
@@ -0,0 +1,77 @@
...
+    </p>
+
+    @php $i = 0; @endphp
+
+    @foreach($events as $eventId => $event)
+
+        @php
+
+            // This is done in this template in order to prevent the jobs table from containing large amounts of event data.
+            $event->event = Proto\Models\Event::find($eventId);
```
jonathanjuursema: Please include withTrashed so the mail does not fail if an event is deleted after dispatching the mail.

jonathanjuursema: Also, Event is aliased so you don't need to include the whole namespace.

PR #980, opened 2019-07-23
Add PIN support to OmNomCom administration. (Fixes #978)

### LINKED ISSUE #978: SumUp integration
Proto recently ordered a SumUp, this of course requiers some integration to replace the current cash payment system. 
short term: we need a way to select payment by pin on the tipcie store. intially a smartphone can be used to communicate with the SumUp
Long term: preferably a single button press would be enough to transfer the payment directly to the sumup, without the use of an additional smartphone

jonathanjuursema:


database/migrations/2019_07_23_195454_add_pin_to_omnomcom.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('orderlines', function (Blueprint $table) {
+            $table->datetime('payed_with_bank_card')->after('payed_with_cash')->nullable()->default(null);
```
JonaMata: Why is this a datetime? Shouldn't it be a boolean?

resources/views/omnomcom/store/show.blade.php
```diff
@@ -721,9 +721,18 @@
...
+        </span>
         @if($store->cash_allowed)
-            <span id="purchase-cash-initiate" class="button inactive">Complete with cash</span>
+            <span id="purchase-cash-initiate" class="button inactive">
+                <i class="fas fa-coins mr-2"></i> Complete with cash
+            </span>
+        @endif
+        @if($store->bank_card_allowed)
+            <span id="purchase-bank-card-initiate" class="button inactive">
+                <i class="fas fa-credit-card mr-2"></i> Complete with PIN
```
JonaMata: PIN might suggest that the payment is completed with a PIN code instead of a bank card.

PR #1037, opened 2019-11-23
(Re)implement basic database seeder for development environment

thijsdortmann:


database/factories/MemberFactory.php
```diff
@@ -0,0 +1,17 @@
...
+
+/** @var $factory Closure */
+$factory->define(Proto\Models\Member::class,
+    function (Faker $faker) {
+        $picktime = $faker->dateTimeInInterval('April 20, 2011', 'now');
+        return [
+            'proto_username' => strtolower(str_random(16)),
+            'created_at' => $faker->dateTime($picktime)->format('Y-m-d H:i:s'),
+            'deleted_at' => (mt_rand(0, 1) === 1 ? null : $faker->dateTimeBetween($picktime, '+1 year')->format('Y-m-d H:i:s')),
+            'is_lifelong' => mt_rand(0, 20) > 15 ? 1 : 0,
```
jonathanjuursema: Maybe you could use these ratios here: https://www.proto.utwente.nl/queries/membership_totals

database/seeds/OtherDataSeeder.php
```diff
@@ -0,0 +1,141 @@
...
+                'user_id' => $users[array_rand($users)],
+                'product_id' => $p->id,
+                'original_unit_price' => $p->price,
+                'units' => $u,
+                'total_price' => $u * $p->price,
+                'created_at' => $d
+            ];
+
+            if (mt_rand(1, 100) == 1) {
+                // Simulate Pay via Cash
```
jonathanjuursema: Payed with card seems more relevant now.

database/seeds/OtherDataSeeder.php
```diff
@@ -0,0 +1,141 @@
...
+        factory(User::class, $n)->create()->each(function ($user) {
+            /** @var $user User */
+            if (mt_rand(1, 5) > 1) {
+                $user->address()->save(factory(Address::class)->make());
+            }
+            if (mt_rand(1, 5) > 1) {
+                $user->bank()->save(factory(Bank::class)->make());
+            }
+            if (mt_rand(1, 5) > 1) {
+                $user->member()->save(factory(Member::class)->make());
```
jonathanjuursema: Members required an address and bank details, so a check for presence of these is necessary. Alternatively, check if a user is member, then add address and bank, otherwise, add address and bank only at random.

database/seeds/OtherDataSeeder.php
```diff
@@ -0,0 +1,141 @@
...
+            }
+        });
+
+        $users = array_merge(array_rand(User::all()->pluck('id')->toArray(), 75), [1]);
+
+        // Create orderlines
+        $n = 10000;
+        echo "Creating $n orderlines" . PHP_EOL;
+
+        $mintime = date('U', strtotime('April 20, 2011'));
```
jonathanjuursema: Perhaps it ie enough to create a number of orderlines in the past 3 months or so. Further back does not add extra.

database/factories/MemberFactory.php
thijsdortmann: I made some approximations based on these ratios.

database/seeds/OtherDataSeeder.php
thijsdortmann: Agreed and changed!

database/seeds/OtherDataSeeder.php
thijsdortmann: I now create address and bank at the same time as the member object, and sprinkle in some additional bank details and addresses for other users.

database/seeds/OtherDataSeeder.php
thijsdortmann: Past year it is :)

PR #1038, opened 2019-11-23
Sort helper reminders by starting date of event. Fixes #994.

### LINKED ISSUE #994: Helper email not ordered on date, but on ID
Zoals genoemd op slack, komt de helper mail binnen, gesorteerd op ID. Het is makkelijker als dit op datum wordt gesorteerd. 

Kusjes

thijsdortmann:
- Sort by property of relation
- Save reminder sent state more efficiently.

### DIFF
--- a/app/Console/Commands/HelperNotificationsCron.php
+++ b/app/Console/Commands/HelperNotificationsCron.php
@@ -46,7 +46,7 @@ class HelperNotificationsCron extends Command
     {
         $users = User::all();
 
-        $handledHelps = [];
+        $handledHelpIds = [];
 
         foreach ($users as $user) {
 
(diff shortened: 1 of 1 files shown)


app/Console/Commands/HelperNotificationsCron.php
```diff
@@ -75,18 +76,24 @@ public function handle()
                             $events[$help->activity->event->id]->help[] = $helpInfo;
                         }
 
-                        $handledHelps[] = $help;
+                        if(!in_array($help->id, $handledHelpIds)) $handledHelpIds[] = $help->id;
                     }
                 }
+
+                unset($committee);
```
jonathanjuursema: Just curious what this is for?

thijsdortmann: As we're doing this for many, many users and many, many committees with many, many relations, I became a bit concerned about memory usage. `unset` simply _unsets_ the variable and lets PHP's garbage collector know this variable is no longer needed. :)

PR #1042, opened 2019-11-24
Add short link service. Fixes #989.

### LINKED ISSUE #989: Shorter link for anonymous email to association board
Heyo, 

Right now the link for the page you can use to send an anonymous email to the association board is https://www.proto.utwente.nl/committee/board/send_anonymous_email. We want to promote this possibility more to our member (for example using the narrowcasting), so it would be nice if there could be a shorter link people can navigate to. Is it possible for you guys to create a 'shortcut link' for this webpage?

xx Ninetje

jonathanjuursema:


resources/views/short_url/edit.blade.php
```diff
@@ -0,0 +1,77 @@
...
+@section('container')
+
+    <div class="row justify-content-center">
+
+        <div class="col-md-4">
+
+            <div class="card mb-3">
+
+                <form method="post" action="{{ route("short_url::edit", ['id' => $url ? $url->id : 'new' ]) }}"
+                      enctype="multipart/form-data">
```
thijsdortmann: I don't think this is necessary, as you're only submitting text

jonathanjuursema: You're right, fixed!

PR #1045, opened 2019-12-01
SAML packages updates so mcrypt is no longer needed.

jonathanjuursema:
Needs to be tested.


jonathanjuursema:
> 
> 
> Let's test this on Tuesday, so we can do a rollback if necessary :) No breaking changes between a 0.1 and 2.1 of `laravel-saml2`?

Haven't looked at the changelog yet. This was what was necessary to make Staging update to the latest version of master for the Google review. This is reminder for me to see if this upgrade doesn't break SAML. :)

app/Http/Controllers/AuthController.php
```diff
@@ -473,7 +473,7 @@ public function passwordSyncPost(Request $request)
     public function startSurfConextAuth()
     {
         Session::reflash();
-        return redirect('saml2/login');
+        return redirect('saml2/surfconext/login');
```
JonaMata: Replace with route

PR #1095, opened 2020-02-18
Implemented society as an option for committees. Fixes #1044.

### LINKED ISSUE #1044: Societies on Website
Here is the issue for working societies into the proto website

Er op komen te staan als societies (niet als commissies) waarbij:
- De leden niet gezien worden als 'actief lid' door lid te zijn van de society
- Er staat bij events die worden gekoppeld aan societies (de 'organized by' unit) dat hij georganiseerd is door een society 
- Niet meetellen voor commissie related achievements want het is geen commissie :p
- Allergielijst moeten ze wel kunnen zien bij events 

There will be a forum on the 10th of december in which our proposel for societies can be reviewed by members. After this, there is a possibility for small changes to still happen in this setup, so postponing starting with this until after the 10th of december is advisable. 

<3 Ninetje

thijsdortmann:


app/Http/Controllers/UserProfileController.php
```diff
@@ -34,9 +34,17 @@ public function show($id = null)
             ->with('committee')
             ->where('user_id', $user->id)
             ->whereNotIn('id', $user->committees->pluck('pivot.id'))
-            ->get();
+            ->get()
+            ->where('committee.is_society', false);
```
JonaMata: If I'm correct less queries are run if the society check is done within the with statement.

app/Models/User.php
```diff
@@ -275,6 +275,24 @@ public function committees()
...
             })
+            ->where('committees.is_society', false)
+            ->where('committees_users.created_at', '<', Carbon::now())
+            ->withPivot(array('id', 'role', 'edition', 'created_at', 'deleted_at'))
+            ->withTimestamps()
+            ->orderBy('pivot_created_at', 'desc');
+    }
+
+    /**
+     * @return mixed Returns all committees a user is currently a member of.
```
JonaMata: Return societies, not committees

app/Models/User.php
```diff
@@ -352,7 +370,10 @@ public function isActiveMember()
                 ->where(function ($q) {
                     $q->whereNull('deleted_at')
                         ->orWhere('deleted_at', '>', date('Y-m-d H:i:s'));
-                })->get()
+                })
+                ->with('committee')
+                ->get()
+                ->where('committee.is_society', false)
```
JonaMata: If I'm correct less queries are run if the society check is done within the with statement.

database/migrations/2020_02_11_203742_create_societies_column_for_committees.php
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('orderlines', function (Blueprint $table) {
```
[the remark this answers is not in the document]
thijsdortmann: Whoops. All hail code reviews.

app/Console/Commands/AchievementsCron.php
```diff
@@ -235,10 +235,14 @@ private function ForeverMember()
         foreach (User::all() as $user) {
             $forever = False;
             foreach (Committee::all() as $committee) {
-                $memberships = CommitteeMembership::withTrashed()->where('user_id', $user->id)->where('committee_id', $committee->id)->get();
+                $memberships = CommitteeMembership::withTrashed()
+                    ->where('user_id', $user->id)
+                    ->where('committee_id', $committee->id)
+                    ->with('committee')
```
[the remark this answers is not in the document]
thijsdortmann: Not strictly necessary, but since it needs to be checked for all of the retrieved objects it should slightly improve performance to do eager loading on this relation.

app/Http/Controllers/CommitteeController.php
```diff
@@ -22,12 +22,12 @@
 class CommitteeController extends Controller
 {
 
-    public function overview()
+    public function overview($showSociety = false)
     {
         if (Auth::check() && Auth::user()->can('board')) {
-            return view('committee.list', ['data' => Committee::orderby('name', 'asc')->get()]);
+            return view('committee.list', ['data' => Committee::where('is_society', $showSociety ? 1 : 0)->orderby('name', 'asc')->get()]);
```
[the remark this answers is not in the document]
thijsdortmann: Fixed :)

app/Http/Controllers/UserApiController.php
```diff
@@ -34,7 +34,12 @@ public function getAddress()
 
     public function getCommittees()
     {
-        return CommitteeMembership::where('user_id', Auth::id())->with('committee')->get();
+        return CommitteeMembership::where('user_id', Auth::id())->with('committee')->get()->where('is_society', false);
```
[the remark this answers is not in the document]
thijsdortmann: I don't quite understand what you mean (can you nest queries in with statements? if so, cool, please show me :)). Since the `where` is behind the `get`, the 'is_society' query is done on the collection, so no queries should be performed.

app/Http/Controllers/UserProfileController.php
thijsdortmann: See above.

app/Models/User.php
thijsdortmann: Fixed :)

app/Models/User.php
thijsdortmann: See above.

app/Http/Controllers/UserProfileController.php
```diff
@@ -34,9 +34,17 @@ public function show($id = null)
...
 
-        return view('users.profile.profile', ['user' => $user, 'pastcommittees' => $pastCommittees]);
+        $pastSocieties = CommitteeMembership::withTrashed()
```
jonathanjuursema: To prevent code-reuse, maybe wrap this query with a "use societies" param?

app/Models/User.php
```diff
@@ -275,6 +275,24 @@ public function committees()
...
+            ->where('committees_users.created_at', '<', Carbon::now())
+            ->withPivot(array('id', 'role', 'edition', 'created_at', 'deleted_at'))
+            ->withTimestamps()
+            ->orderBy('pivot_created_at', 'desc');
+    }
+
+    /**
+     * @return mixed Returns all societies a user is currently a member of.
+     */
+    public function societies()
```
jonathanjuursema: Is it an idea to abstract functions like these into something like get_committees_or_societies (or get_groups) and do something like return get_groups('committees') and get_groups('societies') for code readability and code re-use?

resources/views/committee/include/form-committee.blade.php
```diff
@@ -33,6 +33,19 @@
             </div>
 
             <div class="row">
+                <div class="col-md-6">
+
+                    <div class="form-group">
+                        <label for="public">Committee type</label>
+                        <select class="form-control" id="is_society" name="is_society">
+                            <option value="0" {{ (!$new && $committee->is_society ? '' : 'selected') }}>Regular committee
```
jonathanjuursema: Would just make it `Committee`

routes/web.php
```diff
@@ -213,6 +213,7 @@
         });
 
         Route::get('list', ['as' => 'list', 'uses' => 'CommitteeController@overview']);
+        Route::get('society/list', ['as' => 'list', 'uses' => 'CommitteeController@overview'])->defaults('showSociety', true);
```
jonathanjuursema: Don't we also need a seperate display link?

app/Models/User.php
Jessseee: I think these kinds of issues would be better fixed in a next redesign where we really abstract the societies and committees into something like a 'groups' model.

Jessseee: I think this should be properly fixed when we actually abstract the model of the societies and committees into for example a "group" model.

jonathanjuursema: I see you've marked this as resolved, but don't see any change in this file after the comment has been marked as resolved.

Jessseee: Sorry, I thought this was related to the UserProfileController, instead of the User model. I will fix this as soon as possible.

PR #1166, opened 2020-06-16
Add online good idea board. Fixes #986

### LINKED ISSUE #986: Online Good Idea Board
Heyo Beheer,

Here is an issue about the good idea board we discussed last meeting!
The idea of this tool is that people can post ideas for the association online, similar to the quote corner, with an option to do this anonymously. These ideas can then be "+1"d or "-1"d by other members. 

We as board would like the ability to either remove these ideas when they've been 'up' for too long (or when they are inappropriate), similar to the quote corner, and the ability to 'highlight' ideas when we plan on implementing them / when these ideas have been taken into consideration. 

Thanks already! Love, Board Ninetje <3

JonaMata:


app/Models/GoodIdea.php
```diff
@@ -0,0 +1,28 @@
...
+    public function votes() {
+        return $this->hasMany('Proto\Models\GoodIdeaVote');
+    }
+
+    public function voteScore() {
+        return $this->votes()->sum('vote');
+    }
+
+    public function userVote() {
+        $vote = GoodIdeaVote::where('user_id', Auth::id())->where('good_idea_id', $this->id)->first();
```
jonathanjuursema, line 24: Isn't it better to give a User $user as argument to this function, so you can get the vote for any user?

app/Http/Controllers/GoodIdeaController.php
```diff
@@ -0,0 +1,65 @@
...
+        $ideas = GoodIdea::all();
+        foreach($ideas as $idea) {
+            $idea->delete();
+        }
+    }
+
+    public function vote(Request $request) {
+        $idea = GoodIdea::findOrFail($request->input('id'));
+        $vote = GoodIdeaVote::firstOrCreate(['user_id' => Auth::id(), 'good_idea_id' => $request->input('id')]);
+        $vote->vote = $request->input('voteValue');
```
jonathanjuursema, line 61: No check if someone doesn't input voteValue = 9000? 😁

resources/views/quotecorner/list.blade.php
```diff
@@ -53,7 +53,7 @@
                 },
                 error: function () {
 
-                    window.alert('Something went wrong liking the quote. Please try again.');
+                    console.log('Something went wrong liking the quote. Please try again.');
```
jonathanjuursema, line 56: Maybe also some feedback the user can see?

PR #1184, opened 2020-08-11
Fixes XSS vulnerability. (Fixes #1183)

### LINKED ISSUE #1183: Fix Responsible Disclosure 07/08
Details are once again e-mailed to the committee.

jonathanjuursema:


jonathanjuursema:
I've also checked the database for any users that have an XSS-like string in their diet and reverted them where necessary.

PR #1190, opened 2020-08-19
Redirect to page instead of streaming.

jonathanjuursema:


app/Http/Controllers/UserAdminController.php
```diff
@@ -260,7 +260,7 @@ public function showMemberForm($id)
         $user = User::findOrFail($id);
 
         if ($user->hasSignedMembershipForm()) {
-            return Storage::download($user->member->membershipForm->generatePath(), $user->name . ' membership form.pdf');
```
jonathanjuursema, line 263: I'm not sure why this doesn't work yet, but it seems weird to stream a file we've put to CDN instead of pointing to that file in the first place. :)

Jessseee (approved):
LGTM 👍

PR #1202, opened 2020-09-08
Added full domains for proto.utwente.nl and omnomcom websocket domain

JonaMata:


jonathanjuursema (approved):
LGTM

jonathanjuursema (approved):
LGTM - take into account ProTube URL will most likely change soon due to deprecating metis

PR #1218, opened 2020-10-14
Made user overview admin filterable on user type

Jessseee:
Made user overview admin filterable on user type.


resources/views/users/admin/overview.blade.php
```diff
@@ -9,19 +9,36 @@
...
+                                <input class="form-check-input" type="radio" name="filter" id="members" value="members" @if($filter == 'members') checked @endif>
+                                <label class="form-check-label" for="members"> Members </label>
+                            </div>
+                            <div class="form-check mr-2">
+                                <input class="form-check-input" type="radio" name="filter" id="pending" value="pending" @if($filter == 'pending') checked @endif>
+                                <label class="form-check-label" for="pending"> Pending </label>
+                            </div>
+                            <div class="form-check">
+                                <input class="form-check-input" type="radio" id="all" name="filter" value="" @if($filter == null) checked @endif>
+                                <label class="form-check-label" for="all"> All </label>
```
JonaMata, line 36: Checkboxes hint to the possibility of multiple filters (for example all members and pending members but not normal users) but this doesn't seem possible with the current implementation.

JonaMata (changes requested):
I would change the filter to either support multiple filters, or to use a dropdown or radio input instead

resources/views/users/admin/overview.blade.php
Jessseee: That is why I used radio buttons

app/Http/Controllers/UserAdminController.php
```diff
@@ -26,20 +26,36 @@ class UserAdminController extends Controller
...
+     * @param Request $request
+     * @return \Illuminate\View\View
      */
     public function index(Request $request)
     {
 
         $search = $request->input('query');
+        $filter = $request->input('filter');
+
+        switch ($filter) {
```
Jessseee, line 38: @JonaMata Do you have any comments on the queries, I was wondering if perhaps there is an easier or more elegant method?

resources/views/users/admin/overview.blade.php
JonaMata: nvm then

JonaMata (approved):
LGTM

PR #1219, opened 2020-10-16
Fixed bug making it impossible to create new categories in omnomcom

Jessseee:
Also fixed inconsistent linking to edit page in omnomcom category index table compared to other admin interfaces.


JonaMata (approved):
LGTM

PR #1221, opened 2020-10-22
Fixed the possibility of posting multiple signed membership forms.

Jessseee:
Also updated the membership part of the user admin interface.

### DIFF
--- a/app/Http/Controllers/UserAdminController.php
+++ b/app/Http/Controllers/UserAdminController.php
@@ -54,8 +54,8 @@ class UserAdminController extends Controller
     public function details($id)
     {
         $user = User::findOrFail($id);
-        $memberships['pending'] = Member::withTrashed()->where('user_id', '=', $user->id)->where('deleted_at', '=', null)->where('pending', '=', true)->get();
-        $memberships['previous'] = Member::withTrashed()->where('user_id', '=', $user->id)->where('deleted_at', '!=', null)->get();
+        $memberships = $user->getMemberships();
+
         return view('users.admin.details', ['user' => $user, 'memberships' => $memberships]);
     }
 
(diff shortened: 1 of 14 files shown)


app/Http/Controllers/UserAdminController.php
```diff
@@ -54,7 +54,9 @@ public function index(Request $request)
     public function details($id)
     {
         $user = User::findOrFail($id);
-        return view('users.admin.details', ['user' => $user]);
+        $memberships['pending'] = Member::withTrashed()->where('user_id', '=', $user->id)->where('deleted_at', '=', null)->where('pending', '=', true)->get();
+        $memberships['previous'] = Member::withTrashed()->where('user_id', '=', $user->id)->where('deleted_at', '!=', null)->get();
+        return view('users.admin.details', ['user' => $user, 'memberships' => $memberships]);
```
JonaMata, line 59: It shouldn't happen, but what if a user has multiple non pending non deleted memberships, then only one is shown.

JonaMata (changes requested):

app/Http/Controllers/UserAdminController.php
Jessseee: Should I just add another line for "broken" memberships? 😂

routes/web.php
```diff
@@ -188,12 +188,15 @@
...
-        Route::post('sign', ['as' => 'sign', 'middleware' => ['auth'], 'uses' => 'UserDashboardController@postMemberForm']);
-        Route::get('{id}', ['as' => 'download', 'uses' => 'UserAdminController@showMemberForm']);
-        Route::post('print/{id}', ['as' => 'print', 'middleware' => ['auth', 'permission:board'], 'uses' => 'UserAdminController@printMemberForm']);
-        Route::post('delete/{id}', ['as' => 'delete', 'middleware' => ['auth', 'permission:board'], 'uses' => 'UserAdminController@destroyMemberForm']);
+    Route::group(['prefix' => 'memberform', 'as' => 'memberform::', 'middleware' => ['auth']], function() {
+        Route::get('sign', ['as' => 'sign', 'uses' => 'UserDashboardController@getMemberForm']);
+        Route::post('sign', ['as' => 'sign', 'uses' => 'UserDashboardController@postMemberForm']);
+        Route::group(['prefix' => 'download', 'as' => 'download::'], function() {
+            Route::get('{id}', ['as' => 'new', 'uses' => 'UserAdminController@getNewMemberForm']);
+            Route::get('{id}', ['as' => 'signed', 'uses' => 'UserAdminController@getSignedMemberForm']);
```
JonaMata, line 196: These URLs are identical, these should be changed

JonaMata (changes requested):
LAGTM

JonaMata (approved):
LGTM

PR #1227, opened 2020-10-28
Added option to redirect to job offer directly instead of a description

MaxLiebe:
-  Table joboffers now has a new redirect_url column
- In the Admin 'add job offer' menu there is now an option to enter a redirect url instead of a description
- By clicking on a job offer with a redirect url, you will be redirected to that job offer page. If the job offer does not have a redirect url, it will open the job offer on the Proto site with a description just like before.


Jessseee (dismissed):
LGTM

resources/views/companies/joboffers/edit.blade.php
```diff
@@ -42,8 +42,17 @@
                                    placeholder="Chief Executive Officer" value="{{ $joboffer->title or '' }}" required>
                         </div>
 
-
                         <div class="form-group">
+                            <label>Offer information type</label>
+                            <select class="form-control information_type_selector">
```
JonaMata, line 47: As information_type_selecter specifies a specific field, shouldn't this be an id instead of a class?

resources/views/companies/joboffers/edit.blade.php
```diff
@@ -42,8 +42,17 @@
...
+                                <option value="" @if(!$joboffer || ($joboffer->description == '' && $joboffer->redirect_url == '')) selected @endif disabled>Select a type...
```
JonaMata, line 48: You check the description and redirect_url for empty strings, but their default/disabled values are null.

resources/views/companies/joboffers/edit.blade.php
```diff
@@ -42,8 +42,17 @@
...
+                                </option>
+                                <option value="description" @if($joboffer && $joboffer->description != '') selected @endif>Description</option>
+                                <option value="url" @if($joboffer && $joboffer->redirect_url != '') selected @endif>Redirect URL</option>
```
JonaMata, line 51: Same empty string check, shouldn't this check for null values?

resources/views/companies/joboffers/edit.blade.php
```diff
@@ -42,8 +42,17 @@
...
+                            </select>
+                        </div>
+
+                        <div class="form-group information_type_description">
```
JonaMata, line 55: The information_type_description should probably be an id instead of a class.

resources/views/companies/joboffers/edit.blade.php
```diff
@@ -52,6 +61,22 @@
                             ])
                         </div>
 
+                        <div class="form-group information_type_url">
```
JonaMata, line 64: Here too the information_type_url class should probably be an id

app/Http/Controllers/JobofferController.php
```diff
@@ -105,6 +104,7 @@ public function update(Request $request, $id)
 
         $joboffer->title = $request->title;
         $joboffer->description = $request->description;
+        $joboffer->redirect_url = $request->redirect_url;
```
JonaMata, line 107: Make sure this field and the description field are set to null when an empty string is given

JonaMata (changes requested):
I think it would also be wise to change the joboffer show view/controller to maybe throw a 404 when a joboffer without a description is requested.

JonaMata (changes requested):
LAGTM

.env.docker.example
```diff
@@ -1,98 +0,0 @@
...
-
-PRIMARY_DOMAIN=http://localhost:8080
-REDIRECT_ASSETS_DOMAIN=https://www.proto.utwente.nl
-
-DOCKER_HTTP_PORT=8080
-DOCKER_PMA_PORT=8081
-DOCKER_MAILHOG_PORT=8082
-DOCKER_SQL_PORT=3306
-
-ANALYTICS_URL=analytics.saproto.nl
```
JonaMata, line 98: Please do not delete this file, it a nice friendly file that deserves love

JonaMata (changes requested):
Almost good, but please do not delete precious files

.env.docker.example
MaxLiebe: My past self would have not agreed.

JonaMata (approved):
I like the love this file gets

PR #1231, opened 2020-10-31
Fixed mariadb docker not starting because of mapped volume

JonaMata:
Mariadb handles persistent storage by itself, so a mapped volume is not needed.


Jessseee (approved):
LGTM

PR #1234, opened 2020-11-06
Fixed deleted memberform button removing wrong member object.

Jessseee:
Also fixed wrong orphan check for memberforms in StorageEntry

### DIFF
--- a/app/Http/Controllers/UserAdminController.php
+++ b/app/Http/Controllers/UserAdminController.php
@@ -312,8 +312,9 @@ class UserAdminController extends Controller
             abort(403);
         }
 
-        $user = User::findOrFail($id);
-        $member = $user->member;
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/UserAdminController.php
```diff
@@ -312,8 +312,9 @@ public function destroyMemberForm($id)
             abort(403);
         }
 
-        $user = User::findOrFail($id);
-        $member = $user->member;
+        $member = Member::where('membership_form_id', '=', $id)->first();
+        $user = User::findOrFail($member->user_id);
```
JonaMata, line 316: Isn't there a user relation defined in the member model that can be used here?

JonaMata (changes requested):
LAGTM

JonaMata (approved):
LGTM

PR #1239, opened 2020-11-17
Prevent removal of surname prefixes

JonaMata:


Jessseee (approved):
Unfortunate, but LGTM

PR #1264, opened 2021-03-07
Added the possibility to earn achievements by visiting a webpage.

Jessseee:


database/migrations/2021_03_06_201232_add_url_to_achievement_table.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('achievement', function(Blueprint $table) {
+            $table->boolean('has_page')->after('tier');
+            $table->string('page_name')->nullable()->after('has_page');
```
JonaMata, line 18: If this is used as url, it should be unique. Next to that you'll have to think about url encoding and maybe a seperate name and url field where the url field is highly constrained to a set of allowed characters.

routes/web.php
```diff
@@ -807,6 +807,7 @@
         });
         Route::get('gallery', ['as' => 'gallery', 'uses' => 'AchievementController@gallery']);
     });
+    Route::get('achieve/{achievement}', ['as' => 'achieve', 'uses' => 'AchievementController@achieve']);
```
JonaMata, line 810: There is no auth check in the controller, so you should use the auth middleware here.

resources/views/achievement/admin_includes/edit.blade.php
```diff
@@ -72,4 +108,20 @@
...
+        $('#has_page').on('click', function() {
+            if($(this).is(':checked')) {
+                pageBlock.show();
+                pageBlock.find('input').attr('required', true);
+            } else {
+                pageBlock.hide();
+                pageBlock.find('input').attr('required', false);
+            }
+        });
+    </script>
```
JonaMata, line 126: If the state of checkbox is changed by other means than clicking it, this will not work

resources/views/achievement/admin_includes/edit.blade.php
```diff
@@ -60,6 +60,42 @@
...
+                <label for="has_url">Can be achieved by visiting url</label>
+            </div>
+
+            <div id="achieve_page_block" @if(!$achievement->has_page) style="display: none;" @endif>
+
+                <div class="form-group">
+                    <label for="page_name">Achieve URL</label>
+                    <div class="input-group">
+                        <div class="input-group-prepend">
+                            <span class="input-group-text">saproto.nl/achieve/</span>
```
JonaMata, line 76: maybe try getting that url dynamically. Very future proof if we would ever change domain

JonaMata (changes requested):
LAGTM

resources/views/achievement/admin_includes/edit.blade.php
Jessseee: I checked it and luckily it also works when changing the checkbox with the keyboard. https://stackoverflow.com/questions/11205957/jquery-difference-between-change-and-click-event-of-checkbox#answer-11251498

resources/views/achievement/admin_includes/edit.blade.php
Jessseee: I wouldn't be sure where to get the saproto.nl url, as I'd prefer to use that to keep the url short.

app/Http/Controllers/AchievementController.php
```diff
@@ -128,7 +155,7 @@ public function take($achievement_id, $user_id)
     {
         $achievement = Achievement::find($achievement_id);
         $user = User::find($user_id);
-        if (!$user || !$achievement) abort(404);
+        if (!$user || !$achievement) abort(500, 'User or achievement not found.');
```
JonaMata, line 158: For a not found error use a 404 HTTP error code

JonaMata (changes requested):
LAGTM

JonaMata (approved):
LGTM

PR #1271, opened 2021-03-23
Added Leaderboards!

MaxLiebe:
- Added the option for admins to add, edit and remove leaderboards.
- Every committee can create its own leaderboards with a title, description, icon and most importantly a custom 'point type'.

TODO:
- Allow adding, editing and removing leaderboard entries.
- Link the users in the leaderboard entries back to individual members.
- Add a menu entry so admins don't have to type in the route manually.
- Add a public page where everyone can see the leaderboards.


app/Http/Controllers/LeaderboardController.php
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * Store a newly created resource in storage.
+     *
+     * @param  \Illuminate\Http\Request $request
+     * @return \Illuminate\Http\Response
+     */
+    public function store(Request $request)
+    {
+        $leaderboard = new Leaderboard();
+        $committee = Committee::find($request->input('committee'));
```
Jessseee, line 60: This breaks when there is no committee selected. Should be findOrFail

app/Http/Controllers/LeaderboardController.php
```diff
@@ -0,0 +1,116 @@
...
+    public function update(Request $request, $id)
+    {
+        $leaderboard = Leaderboard::findOrFail($id);
+
+        $leaderboard->name = $request->name;
+        $leaderboard->description = $request->description;
+        $leaderboard->points_name = $request->points_name;
+        $leaderboard->icon = $request->icon;
+
+        $committee = Committee::find($request->input('committee'));
```
Jessseee, line 96: Same here, breaks if no committee is selected

routes/web.php
```diff
@@ -310,6 +310,30 @@
 
     });
 
+    /*
+     * Routes related to leaderboards.
+     */
+    Route::group(['prefix' => 'leaderboards', 'as' => 'leaderboards::'], function () {
```
Jessseee, line 316: Better to just add the auth middleware to the route group.

routes/web.php
```diff
@@ -310,6 +310,30 @@
...
+
+        Route::get('edit/{id}', ['as' => 'edit', 'middleware' => ['permission:board'], 'uses' => 'LeaderboardController@edit']);
+        Route::post('edit/{id}', ['as' => 'edit', 'middleware' => ['permission:board'], 'uses' => 'LeaderboardController@update']);
+        Route::get('delete/{id}', ['as' => 'delete', 'middleware' => ['permission:board'], 'uses' => 'LeaderboardController@destroy']);
+    });
+
+    /*
+    * Routes related to leaderboard entries.
+    */
+    Route::group(['prefix' => 'leaderboards/entries', 'as' => 'leaderboards_entries::'], function () {
```
Jessseee, line 331: Same here, just add the auth middleware to the route group

routes/web.php
```diff
@@ -310,6 +310,30 @@
...
+        Route::get('', ['as' => 'index', 'uses' => 'LeaderboardEntryController@index']);
```
Jessseee, line 332: I don't think it is necessary to index all the leaderboard entries separately from the leaderboards they're associated to

Jessseee (commented):
Some small comments

package-lock.json
```diff
@@ -144,18 +144,18 @@
     "@fortawesome/fontawesome-free": {
```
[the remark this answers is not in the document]
MaxLiebe: This shouldn't be the case, I'll make sure to correct it

routes/web.php
MaxLiebe: I'm pretty sure this route is now obsolete, so I'll remove it in an upcoming commit.

routes/web.php
MaxLiebe: For now, this is can be the case as there is no public page to view the leaderboards yet. However, when the public page goes live this will need to be changed again.

thomasvklink (dismissed):
Looking good!

app/Http/Controllers/LeaderboardController.php
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * Show the form for editing the specified resource.
+     *
+     * @param  int $id
+     * @return \Illuminate\Http\Response
+     */
+    public function edit($id)
+    {
+        $leaderboard = Leaderboard::findOrFail($id);
+        $entries = $leaderboard->entries->sortByDesc('points');
```
JonaMata, line 83: If preferred this could also be handled inside the blade file

app/Http/Controllers/LeaderboardEntryController.php
```diff
@@ -0,0 +1,66 @@
...
+     * @return \Illuminate\Http\Response
+     */
+    public function destroy($id)
+    {
+        $entry = LeaderboardEntry::findOrFail($id);
+        $entry->delete();
+        Session::flash("flash_message", "The entry has been deleted.");
+        return Redirect::back();
+    }
+}
```
JonaMata, line 66: Controllers don't necessarily need to be separated like this, it might be more logical to include this in the leaderboardcontroller

app/Models/Leaderboard.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+
+namespace Proto\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+class Leaderboard extends Model
+{
+    protected $table = 'leaderboards';
```
JonaMata, line 10: This isn't needed as the model name is the singular of the table name

package.json
```diff
@@ -132,7 +132,7 @@
         "./node_modules/bootstrap-slider/dist/css/bootstrap-slider.css",
         "./node_modules/simplemde/dist/simplemde.min.css",
         "./node_modules/select2/dist/css/select2.min.css",
-        "./node_modules/select2-bootstrap4-theme/dist/select2-bootstrap4.scss",
+        "./node_modules/select2-bootstrap4-theme/src/select2-bootstrap4.scss",
```
JonaMata, line 135: This change doesn't seem logical to me, has this been changed on purpose?

resources/views/leaderboards/edit.blade.php
```diff
@@ -0,0 +1,223 @@
...
+                                                    <td>{{ $entry->user->name }}</td>
+                                                    <td style="max-width: 80px">
+                                                        <i class="{{ $leaderboard->icon }}"></i>
+                                                        <span data-id="{{ $entry->id }}" class="le_points" >
+                                                            {{ $entry->points}}
+                                                        </span>
+                                                    </td>
+                                                    <td style="min-width: 60px">
+                                                        <a data-id="{{ $entry->id }}" class="fa fas fa-lg fa-caret-up ml-2 le_increase"></a>
+                                                        <a data-id="{{ $entry->id }}" class="fa fas fa-lg fa-caret-down ml-1 le_decrease"></a>
```
JonaMata, line 129: Am I correct that increasing and decreasing by 1 is the only option to edit points? This could get quite tedious if a lot of people got 10 extra points, maybe also add an input field?

JonaMata (changes requested):
Nice work, only had a few comments

package.json
Jessseee: Yeah, ~/dist/select2-boostrap4.scss doesn't excist which is why the styling of the select2 input fields is kind of weird. But I'll replace it with the .min.css just like the other libraries.

app/Models/Leaderboard.php
Jessseee: I prefer to keep it explisit, just like you can see in the code_quality branch.

JonaMata (approved):
LGTM

PR #1294, opened 2021-04-22
Removed everything related to Slack

Jessseee:


jonathanjuursema:
🪦 RIP

Jessseee:
Yeah we chose to close the Slack for members as no one was using it. We will keep it for the HYTTIAOAc because threading is still nice to have, but we are looking for other alternatives. 😅

PR #1335, opened 2021-08-31
Added membership registration helper role and interfaces

Jessseee:


app/Console/Commands/GenerateRoles.php
```diff
@@ -114,6 +114,12 @@ public function handle()
             $permissions['publishalbums']->save();
             $this->info('Added publishalbums permission.');
         }
+        $permissions['registermembers'] = Permission::where('name', '=', 'signup')->first();
```
JonaMata, line 117: Name should be registermembers as is it is called everywhere else

JonaMata (changes requested):

JonaMata (approved):
LGTM

PR #1337, opened 2021-09-01
Fixed inconsistencies between memberform page and pdf file.

Jessseee:


resources/views/users/includes/membershipform_include.blade.php
```diff
@@ -1,49 +1,33 @@
...
-
 <p>
-    The undersigned is aware of the Bylaws (NL: Statuten) and Rules & Regulations (NL: Huishoudelijk Regelement) of the association and promises to follow them.
-    More information on the before mentioned documents can be found on the <a href="https://wiki.proto.utwente.nl/proto/start" target="_blank">DokuWiki</a>.
+    The undersigned is aware of the
+    <a href="https://wiki.proto.utwente.nl/_media/proto/statues_v3_eng_censor.pdf" target="_blank">Bylaws</a>
+    (NL: <a href="https://wiki.proto.utwente.nl/_media/proto/statuten_v3_nl_censor.pdf" target="_blank">Statuten)</a> and
+    <a href="https://wiki.proto.utwente.nl/_media/proto/rules_and_regulations_s.a._proto_en_.pdf" target="_blank">Rules & Regulations</a>
+    (NL: <a href="https://wiki.proto.utwente.nl/_media/proto/rules_and_regulations_s.a._proto_nl_.pdf" target="_blank">Huishoudelijk Regelement)</a>
+    of the association and promises to follow them.
```
JonaMata, line 17: The Dutch versions of these documents are leading, so only referencing those in parentheses seems a bit odd

JonaMata (changes requested):

resources/views/users/includes/membershipform_include.blade.php
Jessseee: But at the same time the contract is in English, so you have to be clear that the documents you are referring to are in a different language.

JonaMata: I would maybe just add a note that the dutch versions are leading

resources/views/users/includes/membershipform_include.blade.php
```diff
@@ -1,49 +1,33 @@
...
+    <a href="https://wiki.proto.utwente.nl/_media/proto/statues_v3_eng_censor.pdf" target="_blank">Bylaws</a>
+    (NL: <a href="https://wiki.proto.utwente.nl/_media/proto/statuten_v3_nl_censor.pdf" target="_blank">Statuten)</a> and
+    <a href="https://wiki.proto.utwente.nl/_media/proto/rules_and_regulations_s.a._proto_en_.pdf" target="_blank">Rules & Regulations</a>
+    (NL: <a href="https://wiki.proto.utwente.nl/_media/proto/rules_and_regulations_s.a._proto_nl_.pdf" target="_blank">Huishoudelijk Regelement)</a>
+    of the association and promises to follow them. (Please note that the Dutch version of the documents are leading.)
 </p>
-
 <p>
-    In addition, the undersigned promises to pay for any cost incurred as member of the association, including the
-    membership fee, in a timely manner via any of the payment options made available by the association.
```
JonaMata, line 23: Is this line removed on purpose?

Jessseee: Nope, not sure how that happened...

PR #1339, opened 2021-09-01
Fixed local environment referencing static host instead of local files.

Jessseee:
Moved static host url to .env file. **Make sure to change production .env file before merging!**


JonaMata (commented):
@Jessseee What do you think about changing the whole static domain to a .env setting?

Jessseee:
> @Jessseee What do you think about changing the whole static domain to a .env setting?

That is probably better. Just make local the default value.

ysbrandB (approved):
LGTM

PR #1358, opened 2021-09-17
Count pending members seperately in membership totals

JonaMata:


bluesainety (commented):
Code looks good, can't seem to run it in my own dev environment because of laravel log permission issues so I cannot check that!

Jessseee (approved):
LGTM

PR #1363, opened 2021-09-21
Deleted duplications in allergy information lists

bluesainety:
Collection of users is now filtered to only contain unique users based on user id. Duplicated users are not removed in participant checklist etc, let me know if anyone wants to see that added too.


Jessseee (approved):
LGTM

ysbrandB (approved):
LGTM

PR #1364, opened 2021-09-22
Added Xdebug support to docker

JonaMata:


Jessseee (approved):
LGTM. Really needed that.

.env.docker.example
```diff
@@ -5,6 +5,8 @@ FALLBACK_COOKIE_DOMAIN=localhost
 
 DEBUG=true
 
+PHP_IDE_CONFIG=serverName=saproto-debug
```
Jessseee, line 8: Does this take in to account the docker name variable?

Jessseee (changes requested):

.env.docker.example
JonaMata: No, this is the name you give the debug server in your IDE, you just have to make sure that the two match

Jessseee: Maybe we should add that to the README?

JonaMata: wow, so smart

Jessseee (approved):
LGTM

PR #1374, opened 2021-10-03
Disabled adding videos, since it is broken due to our disabled YT API…

JonaMata:
… Key


Jessseee:
Should the video page also be hidden?

Jessseee (approved):
LGTM

PR #1375, opened 2021-10-03
Disabled directadminsync logging

JonaMata:
Disabled logging to stop 'array to string conversion' errors in sentry


Jessseee (approved):
LGTM

PR #1383, opened 2021-10-06
Fixed activity/event sign up end before start

### LINKED ISSUE #1379: Sign up opening must be before closing
The sign-up for events can now open after the sign-up closes. This way you can only put yourself on the backup list, even though the event has no participants.

ysbrandB:
You can now no longer let an activity end before it starts. 
Fixes #1379


Jessseee (approved):
LGTM

PR #1386, opened 2021-10-08
Added .nvmrc and NPM caching in deploy action

Jessseee:
reduces build time from 3 min. to 2 min. when package-lock is not changed.


JonaMata (approved):
LGTM

Issue #1396, opened 2021-10-19
Suggestion dinner form URL


LouisVanMaurik:
When you add a dinner form to the site, it gives suggestions on what to type. For example, the 'Dinner form restaurant:' placeholder text is 'Elat Roma'. It looks like there is also a placeholder text in the URL, however it is not a placeholder text. It is an URL that previously was used as the standard form. Could this text be removed? It is not really that much of a problem, but it sometimes goes wrong if you paste another URL. This is besides the issue that you can't change existing forms.

PR #1405, opened 2021-11-02
Fixed the placeholder link being in the way

ysbrandB:
When making a dinnerform the standard link was not a placeholder and therefore would sometimes be behind the real link when pasted. Is now a proper placeholder. Adds to and closes issue #1396


Jessseee (approved):
LGTM

JonaMata (approved):
LGTM

PR #1411, opened 2021-11-09
Street number instead of housenumber

### LINKED ISSUE #1328: streetnumber instead of house number member signup

ysbrandB:
fixes #1328


Jessseee:
Already fixed in #1337

PR #1421, opened 2021-12-03
Fix dmx_value route and date error for overrides

ysbrandB:
Open up the route to access the JSON data for the smart-XP light system again and disable the laravel protected dates attribute as we do not expect it to be this in our code.


Jessseee (approved):
LGTM

PR #1422, opened 2021-12-05
Ticket urchase button now updates with the right amount

### LINKED ISSUE #1413: When buying tickets the purchase button does not update with the right amount
When you purchase a ticket you get to pick the amount you want and the button beneath it says; "Total: 0.00 Finish Purchase". When you select an amount this button does not update and still says 0.00. The button works and purchases the amount of tickets and euros you should be charged, but it would be nice if it shows the euro amount before purchasing correctly.

ysbrandB:
fixes #1413


Jessseee (approved):
LGTM

PR #1424, opened 2021-12-08
Fix event view for non members

### LINKED ISSUE #1408: event view for non members
When sharing a link to an event on our site, non members just see the location, time and text. So it might not be clear that you have to sign up for such an event. Could a "become a proto member to sign up for this activity" with a "become a member" button be added to the event view for non members?

ysbrandB:
add message for non members to become a member if you need to sign up!
closes #1408

### DIFF
--- a/resources/views/event/display_includes/event_details.blade.php
+++ b/resources/views/event/display_includes/event_details.blade.php
@@ -80,7 +80,17 @@
                 <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i> This event is not organized by S.A. Proto
             </li>
         @endif
-
+        @if (!Auth::check()&&$event->activity !== null)
+            <a href="{{route('becomeamember')}}" class="list-group-item bg-info text-white text-center">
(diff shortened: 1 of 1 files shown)


resources/views/event/display_includes/event_details.blade.php
```diff
@@ -80,7 +80,17 @@
                 <i class="fas fa-fw fa-info-circle" aria-hidden="true"></i> This event is not organized by S.A. Proto
             </li>
         @endif
-
+        @if (!Auth::check()&&$event->activity !== null)
```
Jessseee, line 83: I do not get the difference between the two possible buttons. The text seems to say the same thing. But in case you really want to show different text you could use something like: `{{ isset($event->activity) ? <text> : <other_text> }}` so you don't have to define the same button twice.

Jessseee (changes requested):

resources/views/event/display_includes/event_details.blade.php
ysbrandB: I have now made the division between the buttons more clear and divided the functionality. I want to have a different one reminding you to signup as neccesary. I hope my new commit clears it up 😀

Jessseee (approved):
LGTM

PR #1426, opened 2021-12-10
Added pending member mailinglist

### LINKED ISSUE #1416: Pending member mailing list
Board would like an email list for pending members so they can inform them about and remind them about the fact that just making an account and signing up does not directly make you a member but you need to come visit board and other info.

ysbrandB:
Add a migration for the to_pending mail column in the mail table and add all the necessary code to make the pending mail list work in the controller and necessary views fixes #1416


Jessseee (approved):
LGTM

PR #1433, opened 2021-12-14
Fix/node sass version

JonaMata:


MaxLiebe (approved):
If the pipeline still fails i will be very sad :(

PR #1434, opened 2021-12-15
Orderhistory optimalization

ysbrandB:
Do not let the orderline History retrieve all the users orderlines anymore, only the ones it actually needs.


Jessseee (approved):
LGTM

Jessseee (approved):
LGTM

PR #1436, opened 2021-12-22
Set dmx_values outside the web and inside the api middleware

ysbrandB:
It does not get forced over https anymore (not in a 'web' group) but does get throttled now as it is in an API middleware group

### DIFF
--- a/routes/api.php
+++ b/routes/api.php
@@ -5,7 +5,6 @@ Route::group(['middleware' => ['forcedomain'], 'as' => 'api::'], function () {
     /* Route for smartXp dmx_values (excluded from web and therefore https) */
     Route::get('dmx_values', ['as' => 'dmx_values', 'uses' => 'DmxController@valueApi']);
 
-
     /* Routes related to the General APIs */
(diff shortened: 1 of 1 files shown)


Jessseee (dismissed):
LGTM, hopefully this works and doesn't break anything :)

routes/api.php
```diff
@@ -1,9 +1,13 @@
 <?php
 
 Route::group(['middleware' => ['forcedomain'], 'as' => 'api::'], function () {
+    /* Routes for smartXp dmx_values (excluded from web and therefore https) */
+    Route::group(['middleware' => ['api']], function () {
```
JonaMata, line 5: Routes in routes/api.php automatically get the api middleware assigned, so this line can be removed.

JonaMata (changes requested):
Almost there

routes/api.php
ysbrandB: I had not seen that, it is fixed now

thomasvklink (approved):
LGTM

JonaMata (approved):
LGTM

PR #1437, opened 2021-12-22
Disable the submit button after submit

ysbrandB:
Prevents double submitting a form and therefore doing two php calls

### DIFF
--- a/resources/assets/js/application.js
+++ b/resources/assets/js/application.js
@@ -13,6 +13,18 @@ require('bootstrap-slider')
 require('select2')
 require('fontawesome-iconpicker')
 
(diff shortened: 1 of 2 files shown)


Jessseee (changes requested):
Please only change lines that are necessary, this will make merging with future changes easier. Also, it might be nice to add a [loading icon](https://getbootstrap.com/docs/4.2/components/spinners/) to the submit button, but that is just a suggestion.

Jessseee (changes requested):
please remove the names of the submit buttons you renamed. submit buttons do not need a name field.

PR #1447, opened 2022-01-18
Fixes the need for there to not always be an event in the weekly newsletter

### LINKED ISSUE #1443: Mandatory event when sending the weekly newsletter
Currently an event has to be linked in the weekly newsletter before it can be sent. As there are no(t as much) interesting events, it is a struggle to pick an event which can be promoted in the weekly. It would be nice if this condition could be removed

ysbrandB:
removes the need for an event to be able to send the weekly neweletter, fixes #1443


Jessseee (approved):
LGTM

PR #1449, opened 2022-01-25
Fix zip action zipping entire filesystem

JonaMata:


MaxLiebe (approved):
Jonathan leert hoe file systems werken

PR #1454, opened 2022-02-07
Replace the 9292 api with the rest ovapi

ysbrandB:
The 9292 implementation for the busses on the smartxp screen does not get supported anymore, now replaced by the rest OVapi https://github.com/skywave/KV78Turbo-OVAPI/wiki. The timingpointcodes are for each stop, for each side of the road. Currently hallenweg is discontinued so I chose the closest stop I could get actual data from (langenkampweg)

### DIFF
--- a/app/Http/Controllers/SmartXpScreenController.php
+++ b/app/Http/Controllers/SmartXpScreenController.php
@@ -33,22 +33,20 @@ class SmartXpScreenController extends Controller
      * @param $stop
      * @return object[]
      */
-    public function bus($tpc_id)
+    public function bus($tpc_id, $tpc_id_other_side)
     {
         try {
-            $departures = json_decode(file_get_contents(stripslashes("http://v0.ovapi.nl/tpc/$tpc_id")));
-            return json_encode($departures->$tpc_id->Passes);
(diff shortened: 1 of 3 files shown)


Jessseee (changes requested):
No need for a controller in our website. Everything should be handled client side with javascript AJAX requests.

ysbrandB:
> No need for a controller in our website. Everything should be handled client side with javascript AJAX requests.

Our ajax does not like the http call, returns a weird promise. I did update the way the controller is called.

app/Http/Controllers/SmartXpScreenController.php
```diff
@@ -33,23 +33,19 @@ public function protopenersTimetable()
...
-                'service' => '',
-                'mode' => (object) [
-                    'name' => 'Error in API!',
-                ],
-                'realtimeText' => '',
-                'realtimeState' => '9292',
-                'destinationName' => 'who knows?',
-            ]];
+            return json_encode((object) [
+                'dragon'=>(object) [
```
Jessseee, line 42: Please make this a useful error

Jessseee (changes requested):

app/Http/Controllers/SmartXpScreenController.php
ysbrandB: made it a normal json response where the error is moved to the blade file

Jessseee (approved):
LGTM

PR #1469, opened 2022-02-20
Fixed dasync command

Jessseee:
When receiving the list of all current users from the DirectAdmin API, the list of users is nested in an array with key 'list'. Which means that when comparing to the target list of users a query was generated that was trying to delete a user whose name is equal to an array of all current users.


ysbrandB (approved):
LGTM

PR #1471, opened 2022-02-20
Disabling 1337 soundboard sound cronjob

Jessseee:
Because soundboard is still broken and this cronjob throws an error.


ysbrandB (approved):
LGTM

PR #1474, opened 2022-02-23
Fix mollie payments

Menkveld-24:
Changes on the ticket section of an event at checkout, checkout at the purchase history(balance). 
In order to test the webhook you need to make your pc externally accessible (portforward for example), it also works without but then you need to update payment statuses manually by inspecting the payment. 

There are a few config changes at the omnomcom section, there is a boolean to enable/disable transaction fees and enable methods that will ignore mollie's fees when fees are enabled. 

Ask the mollie test key at Jonathan or Menke


resources/views/event/display_includes/tickets.blade.php
```diff
@@ -147,33 +160,87 @@ class="card-link text-info">
...
             $('.ticket-select').each(function () {
                 total += $(this).attr('data-price') * $(this).val();
             });
-            $('#ticket-total').html(total.toFixed(2))
+            $('#ticket-total').html(total.toFixed(2));
         }
 
     </script>
 
+    @if(!(config('omnomcom.mollie.use_fees') && $only_prepaid) && !(!config('omnomcom.mollie.use_fees') || !$has_prepay_tickets))
```
ysbrandB, line 215: I don't understand why this all needs to be negated, can you either check the not use fees first or make the second half not a negation of a negated or?

resources/views/omnomcom/mollie/list.blade.php
```diff
@@ -79,15 +79,21 @@
                                 </td>
 
                                 <td>
-                                    @if(Proto\Models\MollieTransaction::translateStatus($transaction->status) == 'open')
+                                    {!! Proto\Models\MollieTransaction::translateStatus($transaction->translatedStatus()) == "open" ? '<i class="fas fa-spinner ml-2 text-normal"></i>' : "" !!}
+                                    {!! Proto\Models\MollieTransaction::translateStatus($transaction->translatedStatus()) == "failed" ? '<i class="fas fa-times ml-2 text-danger"></i>' : "" !!}
+                                    {!! Proto\Models\MollieTransaction::translateStatus($transaction->translatedStatus()) == "paid" ? '<i class="fas fa-check ml-2 text-success"></i>' : "" !!}
+                                    {!! Proto\Models\MollieTransaction::translateStatus($transaction->translatedStatus()) == "unknown" ? '<i class="fas fa-question ml-2 text-normal"></i>' : "" !!}
+                                    <span class="label label-default"> - {{ $transaction->status }}</span>
+                                    {{-- Doen de labels wel iets?? --}}
```
ysbrandB, line 87: ? If they dont do anything they can be removed?

ysbrandB (changes requested):
I haven't tested this yet but have found a few things I believe could be left out of the pull-request or done differently

resources/views/event/display_includes/tickets.blade.php
```diff
@@ -147,33 +160,87 @@ class="card-link text-info">
...
+                @endif
             </div>
             @endif
-
         </div>
 
     </form>
 
     <script type="text/javascript" nonce="{{ csp_nonce() }}">
+        console.log("test")
```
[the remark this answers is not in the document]
Menkveld-24: Whoops yes I see the test logs, they indeed need to be removed

app/Http/Controllers/MollieController.php
```diff
@@ -17,43 +18,91 @@
...
-        $orderlines[] = $orderline->id;
+            if ($selected_method->count() === 0) {
+                return Redirect::back()->with('flash_message','The selected payment method is unavailable, please select a different method');
+            }
 
-        $transaction = self::createPaymentForOrderlines($orderlines);
+            $selected_method = $selected_method->first();
+        
+            if (
+                $total < floatval($selected_method->minimumAmount->value) ||
```
MaxLiebe, line 98: Float conversion here does not seem like a good idea. This could lead to rounding errors causing ambiguous behavior. If possible, avoid it. Although for this maybe the OrderLine models should be migrated to use integer types instead of floats. For being a rarely used feature, it could be okay.

ysbrandB (approved):

MaxLiebe (dismissed):
LGTM

ysbrandB (approved):
LGTM

Jessseee (approved):
LGTM

PR #1480, opened 2022-02-25
Added concurrency to GitHub deploy action

Jessseee:
This will cancel any already running deploy actions to stop them from accidentally overwriting each other. See [github actions docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow) on concurrency for more info.


ysbrandB (approved):
LGTM

PR #1492, opened 2022-03-04
Fixed error in personal purchase history

Jessseee:
Purchase history gave an error when there were no orderlines for the current month


ysbrandB (approved):
LGTM

PR #1493, opened 2022-03-04
Fixed error in withdrawal when user (somehow) has no bank

Jessseee:
Withdrawals broke when a deleted user was on there.


ysbrandB (approved):
LGTM

PR #1498, opened 2022-03-07
Added possibility for admins to disable a users 2fa

Jessseee:
Should board be able to disable a user's 2fa or should only sysadmins be allowed to?


ysbrandB (approved):
LGTM

PR #1499, opened 2022-03-08
Show secret activities on the homepage when signed up

ysbrandB:
Show the secret events also in upcoming events on the homepage when signed up.

### DIFF
--- a/resources/views/website/layouts/macros/upcomingevents.blade.php
+++ b/resources/views/website/layouts/macros/upcomingevents.blade.php
@@ -28,27 +28,31 @@
     </div>
     <div class="card-body">
 
-        @php($events = Proto\Models\Event::where('secret', false)->where('is_featured', false)->where('end', '>=', date('U'))->orderBy('start')->limit($n)->get())
+        @php($events = Proto\Models\Event::where('is_featured', false)->where('end', '>=', date('U'))->orderBy('start')->limit($n)->get())
 
         @if(count($events) > 0)
 
             @foreach($events as $key => $event)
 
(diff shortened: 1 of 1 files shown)


resources/views/website/layouts/macros/upcomingevents.blade.php
```diff
@@ -28,27 +28,31 @@
...
 
-        @php($events = Proto\Models\Event::where('secret', false)->where('is_featured', false)->where('end', '>=', date('U'))->orderBy('start')->limit($n)->get())
+        @php($events = Proto\Models\Event::where('is_featured', false)->where('end', '>=', date('U'))->orderBy('start')->limit($n)->get())
 
         @if(count($events) > 0)
 
             @foreach($events as $key => $event)
 
-                @include('event.display_includes.event_block', ['event'=> $event])
+                @if(!$event->secret || (Auth::check() && ($event->activity && $event->activity->isParticipating(Auth::user()))))
```
Jessseee, line 37: Maybe also add a condition for the board permission so they can always see the hidden activities like in the agenda.

Jessseee (changes requested):

resources/views/website/layouts/macros/upcomingevents.blade.php
ysbrandB: I have purposefully not done this as the board may want to show their homepage without spoiling any secret events they are not going to. They can instead still see them all at the overview

Jessseee (approved):
LGTM

PR #1505, opened 2022-03-14
Checkmark with tooltip on the achievement block include when achieved

ysbrandB:
you can now see on the achievement overview page which achievements you have achieved!


MaxLiebe (commented):
LGTM and +1 for achievedMent

Jessseee (approved):
LGTM

PR #1508, opened 2022-03-15
Import Quagga in application js

JonaMata:


ysbrandB (approved):
LGTM 10/10 for the amount of lines changed per commit :)

PR #1511, opened 2022-03-20
Fix/photo album back button

### LINKED ISSUE #1445: Pressing the back arrow while looking at a photo in photo albums puts you back to the first page, instead of the page you were on.

ysbrandB:
The backbutton in an photoalbum now goes back to the page in the album the photo is actually on. Fixes #1445.


Jessseee (approved):
LGTM

PR #1513, opened 2022-03-20
Achievement for the 1st and 100th activity (nTh)

### LINKED ISSUE #1053: Achievement proposal: First/100th activity
Hey beheer,

Would it be possible to have an achievement for when someone signs themselves up (and stays signed up until the event has actually happened, aka they are actually present) for their first activity? And consequently, the same kind of thing for their 100th activity? 

<3 board 9.0

ysbrandB:
**IMPORTANT**
Before merging the achievements need to get added and the achievement ID's added to line 72 and 73!
Also an extra test wouldn't hurt.
resolves #1053.


Jessseee (dismissed):
LGTM

PR #1514, opened 2022-03-21
percentage of activities attended

### LINKED ISSUE #1052: Achievement proposal: Participation Trophee
Hey have you tried, 

We have a proposal for a new achievement: the participation trophee! This is an achievement that is given when a member has attended a certain % or Proto activities in a month. This would then be a % of all proto activities that actually has a sign up. 
We like the idea of having achievements for 25%, 75% and 100%, if it isn't too much work to make the achievement for different percentages. What do you guys think about this? 

<3 board 9.0

ysbrandB:
**IMPORTANT
Before merging the achievements need to get added and the achievement ID's added to line 72 and 73!
Needs to be properly tested!**

Provide an achievement for a percentage of activities attended in a month!
resolves #1052

### DIFF
--- a/app/Console/Commands/AchievementsCron.php
+++ b/app/Console/Commands/AchievementsCron.php
@@ -6,8 +6,11 @@ use Carbon;
 use Illuminate\Console\Command;
 use Proto\Models\Achievement;
 use Proto\Models\AchievementOwnership;
+use Proto\Models\Activity;
+use Proto\Models\ActivityParticipation;
 use Proto\Models\Committee;
 use Proto\Models\CommitteeMembership;
+use Proto\Models\Event;
 use Proto\Models\Member;
 use Proto\Models\OrderLine;
 use Proto\Models\Product;
@@ -66,6 +69,9 @@ class AchievementsCron extends Command
         $this->giveAchievement($this->ForeverMember(), 38);
         $this->giveAchievement($this->GoodHuman(), 53);
         $this->giveAchievement($this->IAmNoodle(), 54);
+        $this->giveAchievement($this->percentageActivities(25), 55);
+        $this->giveAchievement($this->percentageActivities(75), 56);
+        $this->giveAchievement($this->percentageActivities(100), 57);
 
         $this->info('Auto achievement gifting done!');
     }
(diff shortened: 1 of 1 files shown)


app/Console/Commands/AchievementsCron.php
```diff
@@ -513,4 +519,26 @@ private function bigKid()
...
+
+                    $possibleActivities = Event::where('end','>',Carbon::now()->subMonth()->valueOf())->where('end','<',Carbon::now()->valueOf())->where('secret','=', '0')->has('activity')->count();
+
+                    if (floor($CountEventsParticipated / $possibleActivities * 100) >= $percentage) {
+                        $selected[] = $user;
+                    }
+                }
+            }
+            else {
+                $this->info('Its not the first of the month! Cancelling '.$percentage.' of Activities this month!');
```
Jessseee, line 540: I don't quite understand this message. What does "Canceling `percentage` of activities this month!" mean?

ysbrandB: I have updated the cancellation message to be more clear.

app/Console/Commands/AchievementsCron.php
```diff
@@ -532,4 +535,26 @@ private function nThActivity($activityAmount) {
...
+            if (Carbon::now()->day == 1) {
+                $users = User::all();
+                foreach ($users as $user) {
+                    $participated = ActivityParticipation::where('user_id', $user->id)->pluck('activity_id');
+                    $activities = Activity::whereIn('id', $participated)->pluck('event_id');
+                    $CountEventsParticipated = Event::whereIn('id', $activities)->where('end','>',Carbon::now()->subMonth()->valueOf())->where('end','<',Carbon::now()->valueOf())->where('secret','=', '0')->count();
+
+                    $possibleActivities = Event::where('end','>',Carbon::now()->subMonth()->valueOf())->where('end','<',Carbon::now()->valueOf())->where('secret','=', '0')->has('activity')->count();
+
+                    if (floor($CountEventsParticipated / $possibleActivities * 100) >= $percentage) {
```
Jessseee, line 550: Maybe we want a minimum number of possible activities? otherwise you could get a divide by 0 error.

Jessseee (changes requested):

ysbrandB:
Due to the rewrite of the achievement cron this had to be rewritten so now closed.

PR #1516, opened 2022-03-21
Show amount of currently signed up members to an activity in the include

### LINKED ISSUE #1174: participant number visibility
Heyo,  
perhaps it would be cool to have the number of participants signed up for an activity visible on the main page. 
Cheers

ysbrandB:
resolves #1174

### DIFF
--- a/resources/views/event/display_includes/event_block.blade.php
+++ b/resources/views/event/display_includes/event_block.blade.php
@@ -86,13 +86,8 @@ Auth::check() && (($event->activity && $event->activity->isParticipating(Auth::u
             @if($event->activity && $event->activity->users->count()>0)
                <br>
                <span>
-                   @if($event->activity->isFull())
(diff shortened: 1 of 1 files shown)


resources/views/event/display_includes/event_block.blade.php
```diff
@@ -83,6 +83,19 @@
                 </span>
             @endif
 
+            @if($event->activity && $event->activity->users->count()>0)
+               <br>
+               <span>
+                   @if($event->activity->isFull())
+                       <i class="fas fa-user-alt fa-fw" style="color: orange;" aria-hidden="true"></i>
+                       Full!
```
Jessseee, line 91: Maybe add something to incentivize members to sign up for the backup list if it is available? Otherwise people might see that an event is full and not bother looking at the event anymore.

ysbrandB: I agree that that would be a nice addition but do not have a clear solution for it. Do you have any input?

PR #1519, opened 2022-03-23
Http to https automatically redirect

Menkveld-24:
Docker nginx is modified to automatically redirect non-https requests to https so you don't have to type https://... every time

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,9 +39,12 @@ services:
     ports:
       - 127.0.0.1:${DOCKER_HTTP_PORT}:443
     volumes:
-      - ./docker/nginx/nginx.conf:/etc/nginx/nginx.conf
+      - ./docker/nginx/nginx.conf:/etc/nginx/templates/nginx.conf.template
       - ./docker/nginx-log:/var/log/nginx
       - .:/var/www
+    environment:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -39,9 +39,12 @@ services:
     ports:
       - 127.0.0.1:${DOCKER_HTTP_PORT}:443
     volumes:
-      - ./docker/nginx/nginx.conf:/etc/nginx/nginx.conf
+      - ./docker/nginx/nginx.conf:/etc/nginx/templates/nginx.conf.template
```
Jessseee, line 42: Why are you changing this directory?

Menkveld-24: I didn't want to hardcode port 8080 for the redirect and I can't use the incoming port (it thinks it comes from 443). So I wanted to use the same port variable as in the .env. But nginx is unable to resolve the set environment variable. Here comes envsubst, it is built into nginx (see https://hub.docker.com/_/nginx at "Using environment variables in nginx configuration"), so nginx substitutes the variable and then replaces the file. But in order to do that the file first needs to be in the dedicated directory for this

Jessseee: ohboi, well if it works...

Jessseee (approved):
LGTM

Issue #1518, opened 2022-03-22
User statistics


ysbrandB:
I think it might be fun to have a page to show user statistics.
For instance:
-Total signups
-Total money spent
-Total products purchased

PR #1522, opened 2022-03-25
Added user statistics to profile

ysbrandB:
Added some fun basic user statistics on your profile page. 
This is only visible to the authenticated user.
Solves #1518


Jessseee (commented):
Maybe add a disclaimer that only the user themselves can see these statistics, to prevent any confusion. As the profile is normally a 'public' page.

Jessseee (approved):
LGTM

PR #1531, opened 2022-04-05
Make tooltip orange when on backuplist

ysbrandB:
Change the tooltip on the included events-block from green to orange when on the backup list and announce that you are on the backuplist with the tooltip. Avoids confusion.

### DIFF
--- a/resources/views/event/display_includes/event_block.blade.php
+++ b/resources/views/event/display_includes/event_block.blade.php
@@ -28,10 +28,13 @@ Auth::check() && (($event->activity && $event->activity->isParticipating(Auth::u
                    data-toggle="tooltip" data-placement="top" title="This is a featured activity!"></i>
             @endif
             @if($event->activity && Auth::check())
-                @if($event->activity->isParticipating(Auth::user()))
-                    <i class="fas fa-check text-primary fa-fw" aria-hidden="true"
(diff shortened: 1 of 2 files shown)


Jessseee (changes requested):
I think a helper method in the activity model, like isParticipating but for the backuplist, would better than doing database queries in the blade file.

ysbrandB:
> I think a helper method in the activity model, like isParticipating but for the backuplist, would better than doing database queries in the blade file.

The logic is now moved to the controller

resources/views/event/display_includes/event_block.blade.php
```diff
@@ -28,10 +28,13 @@
                    data-toggle="tooltip" data-placement="top" title="This is a featured activity!"></i>
             @endif
             @if($event->activity && Auth::check())
-                @if($event->activity->isParticipating(Auth::user()))
-                    <i class="fas fa-check text-primary fa-fw" aria-hidden="true"
-                       data-toggle="tooltip" data-placement="top" title="You are participating!"></i>
-                @endif
```
Jessseee, line 34: This will now always show the user as participating, right?

Jessseee (changes requested):

Jessseee (changes requested):

Jessseee (approved):
LGTM

PR #1543, opened 2022-04-16
Fixed the images being cached for only the default 5 minutes

ysbrandB:
We only cached images for 5 minutes on disk due to a missing argument in the image::cache.
Now we will store them for way longer (about two months) on disk.


Jessseee (commented):
_**Really?!**_

Jessseee (changes requested):
I think the event block should also change.

ysbrandB:
> I think the event block also needs a change.

I do not think this is the case, yes these will have different sizes but it is nice to have them bigger than 400x400 and they will be cached anyways by getting that size via the function. This was always already a different size.

PR #1560, opened 2022-04-28
Fixed selectable photo calling old thumb function

ysbrandB:
Missed because of the magic or implicit call, now fixed


Jessseee (approved):
LGTM

PR #1571, opened 2022-05-02
Added possibility for board to sign up people for event when event has 0 participants

JonaMata:


Jessseee (approved):
LGTM, thought I already fixed this...

PR #1575, opened 2022-05-03
Fixed dasync not comparing current and target forwarders list correctly

Jessseee:
and fixed .env not being loaded correctly by clearing config cache after deploy


ysbrandB (approved):
LGTM

PR #1578, opened 2022-05-03
fixes the ability for photos to overflow on the photopage

ysbrandB:
Fixed the photos overflowing on the photopage


Jessseee (approved):
LGTM

PR #1586, opened 2022-05-09
New message on the omnomcom for soccer cards

ysbrandB:
Add a message to the omnomcom when a purchase is finished to show how many staff football card you may take per purchase as requested by the sportcie


JonaMata (approved):
LGTM

PR #1587, opened 2022-05-09
hopefully fix tipcie store not accessible

MaxLiebe:


Jessseee (changes requested):
Not even close... These are the routes to the tipcie order history. The problem is in the OmNomController. It is still using `Auth::user()->can` to check multiple permissions. I somehow missed this before.

PR #1596, opened 2022-05-11
Show an open lock icon when a user can signup for an event

ysbrandB:
When the user can signup an open lock will now be shown in the event include block


Jessseee (approved):
LGTM

PR #1598, opened 2022-05-12
Add clickable links for the events in an email and proto logo

ysbrandB:
Add clickable links for the events in an email and a link to the homepage in the proto logo in an email.
The actual mail needs to be still tested.
- [ ] Test the actual sending of an email because mailchimp doesn't work for me

### DIFF
--- a/resources/views/emails/template.blade.php
+++ b/resources/views/emails/template.blade.php
@@ -51,11 +51,11 @@
     </tr>
 </table>
 
-<div style="text-align: center;">
+<center>
     <a href="{{route('homepage')}}">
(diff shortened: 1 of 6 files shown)


resources/views/emails/template.blade.php
```diff
@@ -51,9 +51,11 @@
     </tr>
 </table>
 
-<center>
+<div style="text-align: center;">
```
Jessseee, line 54: this will break in some email clients...

Jessseee (changes requested):

ysbrandB (commented):
Reverted the center styling

PR #1602, opened 2022-05-12
Add an extra control bar at the bottom of a quotepage with >10 quotes

ysbrandB:
This way you do not have to scroll all the way to the top again when done with reading this page


Jessseee (approved):
LGTM

PR #1606, opened 2022-05-15
Is Alfred There message

ysbrandB:
Add the ability for Alfred to put a custom message on the page when he is away or when he resets his whereabouts. The statuses have remained the same to keep backwards compatibility with the api.


Jessseee (approved):
LGTM

PR #1607, opened 2022-05-15
Improved performance and readability of achievements cron

Jessseee:
Still needs a fix for the _You're Special_ achievement.


ysbrandB (dismissed):
LGTM

PR #1614, opened 2022-05-18
Prevent users being plucked from the backup-list prematurely

ysbrandB:
This happened when there were too many users on the event (accidental duplicates from our website being overloaded).
If there were more users than places and one got deleted there would be one moved from the backuplist even if there was no space for them.


Jessseee (approved):
LGTM

PR #1615, opened 2022-05-18
Add the Participation Trophee function

ysbrandB:
Add the automated function for 25, 50 and 75% of all signups attended.
Also added them on the site with the right ID's :)


Jessseee (approved):
LGTM

PR #1635, opened 2022-07-05
Clarify the status of an withdrawal

ysbrandB:
When users could not close their account due to 'pending' withdrawals they could not see this reason anywhere. Now I have clarified the message and added the status to the withdrawals so the user can actually figure out why they can not close their account.

### DIFF
--- a/resources/views/omnomcom/orders/includes/recent-payments.blade.php
+++ b/resources/views/omnomcom/orders/includes/recent-payments.blade.php
@@ -9,17 +9,21 @@
         <ul class="list-group list-group-flush">
 
             @foreach($user->withdrawals(6) as $withdrawal)
-                <li class="list-group-item">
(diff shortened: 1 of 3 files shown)


resources/views/omnomcom/orders/includes/recent-payments.blade.php
```diff
@@ -9,17 +9,21 @@
         <ul class="list-group list-group-flush">
 
             @foreach($user->withdrawals(6) as $withdrawal)
```
Jessseee, line 11: Why is that 6 there?

ysbrandB: The 6 is because this is the recent withdrawal form and thats how the function takes in the limit

Jessseee (approved):
LGTM

PR #1637, opened 2022-07-16
Update/photo system

ysbrandB:
Updates for the photo-system
- Made each photo have five file entry's of differing resized quality (and implemented the migration for it)
- Made public photos not go through an controller anymore but instead put them on the public storage
- Made the photopage Ajax (non-reloading)
- Made photoManager redundant
- Added progressive loading to the photo page
- Properly implemented the deletion of albums
- Fixed the erorr message in the upload tool so it does not disappear when one photo uploads successfully
- Added a download button for the original file (also on key arrow-down)
- Made the photo like toggle via a get-request (on key arrow-up)
- Fixed the order of photos and who can view private photos/albums
- Added the ability for the server to automatically add a watermark


ysbrandB:
Figured there is still more work to do so will reopen when ready

PR #1639, opened 2022-07-21
Feature/integrated dinnerform orderlines

### LINKED ISSUE #1430: Replacing the orderline and dinner form
At the moment: I need to make a google form. Add a dinner form on the website. (Then I order what was in the dinner form but that cant be done automatically). But after that, I need to write down every order of every person with the price in a orderline wizard.

If there would be a feature that I can make a dinner form, which already has a google form build in, I wouldn't have to make google form. AND I wouldn't have to type it over as an orderline. If you want to make my life really easy, it would be great if I can add the price of the bill as an after thought. Maybe you could make it check automatically if the bill corresponds with the sum of the prices people filled in. This creates one problem: helpers should have helpers discount. So on the google form, there should be option to say you are a helper. If you really want to make it fancy, you can maybe link dinner form to events. Then you could already see if people are helping during a event. So only if they are helping you have that option. If this is to much, it would be great that there is a message somewhere (probably at the place where you fill in the price of the bill) which says who filled in that they are a helper.

I know this is a bigggg feature, but I think it would be so handy.

ysbrandB:
Restarted the integrated dinnerform again from scratch. Implements and fixes #1430.
This means that the dinnerforms now fully go through our own website, can be linked to an event to get the helpers and afterwards using the TakeAway Dinner product get converted to orderlines. 

To test:
- Reseed DB to get the new product of of the live website
- Run Php artisan optimize:clear to get the change to the omnomcom config


app/Models/Dinnerform.php
```diff
@@ -57,4 +60,48 @@ public function hasExpired()
...
+    public function event()
+    {
+        return $this->belongsTo('Proto\Models\Event', 'event_id');
+    }
+
+    public function totalAmount() {
+        return $this->orderlines()->sum('price');
+    }
+
+    public function totalAmountwithHelperDiscount() {
```
Jessseee, line 80: Helper discount should be an amount not a percentage.

Jessseee (changes requested):

app/Models/Dinnerform.php
ysbrandB: Fixed :)

PR #1640, opened 2022-07-22
Fixed the isAlfredthere timer

ysbrandB:
Added a global timerlist and a check to check if the countdown-timer is already started.


MaxLiebe (approved):
LGTM

PR #1645, opened 2022-07-31
Improve orderline history, menuitems and events

ysbrandB:
Drastically reduce the amount of DB calls the pages are doing


Jessseee (approved):
LGTM

PR #1646, opened 2022-07-31
Add clockwork to composer

ysbrandB:
https://underground.works/clockwork/
https://localhost:8080/clockwork/app#
To help profile what DB calls and models laravel is using in dev environment,
gets automatically discovered by laravel. Can extend the profiler of xDebug even more.


Jessseee (approved):
LGTM

PR #1649, opened 2022-08-08
add Remaining outstanding amount

### LINKED ISSUE #1648: Extra information about how much money is in activities you are signed up for but are not closed (in the withdrawal) yet
This year for instance prototrip took very long to close, and when it did seemingly out of nowhere a lot of money was subtracted out of my account. It would be nice to see how much money you owe in past events that are not closed yet, so you know how much money to hold aside 😅.

ysbrandB:
Fixes #1648.
Adds the information about how much money you owe in activities that you are signed up for but are not closed yet.

### DIFF
--- a/app/Http/Controllers/OrderLineController.php
+++ b/app/Http/Controllers/OrderLineController.php
@@ -2,13 +2,16 @@
 
 namespace Proto\Http\Controllers;
 
+use Activities;
 use Auth;
 use Carbon;
 use DB;
 use Exception;
+use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
 use Illuminate\View\View;
+use Proto\Models\Activity;
 use Proto\Models\FailedWithdrawal;
 use Proto\Models\OrderLine;
 use Proto\Models\Product;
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/OrderLineController.php
```diff
@@ -60,6 +62,11 @@ public function index($date = null)
             }
         }
 
+        $outstanding = Activity::whereHas('users', function (Builder $query) {
```
Jessseee, line 65: Could you also add the remaining amount of other products?

Jessseee (changes requested):

app/Http/Controllers/OrderLineController.php
ysbrandB: I just tried that but think it is confusing because they will be under the 'next withdrawal' amount. This amount will only be things that will not be in the next withdrawal which the other products always are.

ysbrandB: The only thing that is not under the next withdrawal en will be subtracted later are unclosed events which will be in this amount.

Jessseee: Fair

Jessseee (approved):
LGTM

PR #1667, opened 2022-09-13
fix event validation

### LINKED ISSUE #1254: Updating event adding issue
Hey guys, 

I noticed that when you create a new event and try to submit it with faulty content (an event ending before it starts, et cetera) the whole event you tried to add is cleared. This is quite annoying as promotional texts sometimes are not written down anywhere else and have to be rewritten if this happens. You also have to fill in every box again, which is quite tedious. It would be lovely if the error that now shows before wiping everything, has the option shows to change the faulty data instead of having to do everything again (or if the error shows only the faulty data is wiped instead of everything). 

Thanks guys <3

ysbrandB:
Implement event validation in a request and make it so the error gets shown nicely and the information gets repopulated.
Fixes #1254

### DIFF
--- a/app/Csp/Policies/ProtoPolicy.php
+++ b/app/Csp/Policies/ProtoPolicy.php
@@ -3,6 +3,7 @@
 namespace Proto\Csp\Policies;
 
 use function Sentry\captureException;
+
 use Spatie\Csp\Directive;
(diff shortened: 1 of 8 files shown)


Jessseee (approved):
LGTM

PR #1673, opened 2022-09-14
Fixed the ability to close withdrawals for users without a bank

ysbrandB:


app/Http/Controllers/WithdrawalController.php
```diff
@@ -385,7 +385,9 @@ public static function close(Request $request, $id)
         }
 
         foreach ($withdrawal->users() as $user) {
```
Jessseee, line 387: This foreach loop can be removed as we are not editing the bank object anymore after removing the is_first boolean.

Jessseee (changes requested):

app/Http/Controllers/WithdrawalController.php
ysbrandB: Yeah you're right, fixed

Jessseee (approved):
LGTM

PR #1680, opened 2022-09-19
Fixed the user updating when editing

ysbrandB:
Whoops missed this when only testing with my own account


Jessseee (approved):
Whoops, LGTM

PR #1691, opened 2022-09-24
Optimize/event overview page

ysbrandB:
Prevented loading of ALL activities we have ever done when visiting the overview page lol, now restricted to only the year you're viewing. Cuts down on half the memory used by the page and 2/3's of the loading time :)


Jessseee (approved):
LGTM

Issue #1687, opened 2022-09-22
Add option to make a dinnerform that is not visible on the homepage


ysbrandB:
Could be handy for only specific events or if something went wrong

Issue #1681, opened 2022-09-19
Dinnerform on event page


Jessseee:
It could be useful to also display the dinnerform card on the related event.

PR #1693, opened 2022-09-25
Features/dinnerform additions

ysbrandB:
Implements #1681 and #1687.
If a dinner form is linked to an event the event page will now also show the dinner form.
Also added the option to hide a dinner form from the home page.


Jessseee (approved):
LGTM

PR #1718, opened 2022-10-06
fixed birtdays not showing up on the homepage

ysbrandB:
Actually checked for true.
Jesse did an oopsy el mayonais, checked the pull request once more and this was the only occurence :)


Jessseee (approved):
Whoops

PR #1721, opened 2022-10-07
fix/withdrawal accounts date

ysbrandB:
fixed date now being a carbon object


MaxLiebe (approved):
lookie good to me

Jessseee (commented):
I think using `Carbon::parse($orderline->created_at)` would be nicer

PR #1734, opened 2022-10-13
Made the passwordstore copy feedback always visible

### LINKED ISSUE #1701: A notification if you copy a username or password
Now it is not clear when you copy a username or password in the password. There is no notification/pop-up

JonaMata:
Fixes #1701

### DIFF
--- a/resources/assets/js/application.js
+++ b/resources/assets/js/application.js
@@ -25,6 +25,7 @@ if (discordOnlineCount) {
 
 // Enables tooltips elements
 import { Tooltip } from 'bootstrap'
+global.Tooltip = Tooltip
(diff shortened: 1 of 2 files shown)


resources/assets/js/application.js
```diff
@@ -25,6 +25,7 @@ if (discordOnlineCount) {
 
 // Enables tooltips elements
 import { Tooltip } from 'bootstrap'
+global.Tooltip = Tooltip
```
Jessseee, line 28: I don't think making Tooltip global is a great solution for this problem.

Jessseee (changes requested):
DNLGTM

MaxLiebe (approved):
LGTM

PR #1737, opened 2022-10-16
Added information on additional coding tools to the README

Jessseee:


ysbrandB (commented):
I think it would also be nice to add a bit for clockwork and how to integrate that with xdebug

PR #1740, opened 2022-10-19
Fixed the dataseeder for non HYTTIOAOAc members

Jessseee:
The dataseeder gave an ambiguous error when a user is not a member of the HYTTIOAOAc or has not signed an NDA.


ysbrandB (approved):
LGTM

PR #1743, opened 2022-10-20
fixed events defaulting to public when saved and saved again

ysbrandB:
accidently making events public because you changed something and did not double check the visibility is not fun. Now fixed


Jessseee (approved):
LGTM

PR #1749, opened 2022-10-25
Added the possibility to set a percentage discount for a dinnerform

### LINKED ISSUE #1747: Discount dinner form
Hello my lovely beheer friends,
I hope you have missed me, it has been a while since I posted my last issue. Since this issue is all about food, here is an inspirational food quote:

"Food may be essential as fuel for the body, but good food is fuel for the soul"

And besides that it is important that it is cheap because **student life**. That is why our dear external made a deal with Oh my Grill that we get a 12% discount, however this means I now have to change all the orderlines manually to add the discount when I am processing a dinner form. This results in sad penny noises. Would it be possible to have a feature on the dinner form page where you can implement a discount automatically? Then the sad penny noises turn into very happy penny noises x

Kusjes!
Your penny

Jessseee:
Also improved the dinnerform UI and refactored the related models and controllers. Fixes #1747.


ysbrandB (approved):
LGTM

PR #1751, opened 2022-10-28
Fix spelling mistake clientside

Wilkuu-2:
Past tense of "pay" is "paid", not "payed"

This pull request only fixes this mistake for the client,
but this mistake is also present in the code and the database.
Fixing that requires a migration.


Jessseee (approved):
LGTM

PR #1753, opened 2022-10-28
fixed the performance of nightMode by recreating it using css keyframe animations

ysbrandB:
title says it all again :)


Jessseee (dismissed):
LGTM

PR #1754, opened 2022-10-31
check if there are actually helpingcommittee instances

ysbrandB:
fixes the sentry view exception where a dinneform is present but not a helpingcommittee instance


MaxLiebe (approved):
LGTM

PR #1755, opened 2022-10-31
add a back to calendar button and show the publication time in the card for events

ysbrandB:
show the publication time and add a back to calendar button on events, also clarify that the save button only saves the event details.


MaxLiebe (approved):
GLTM

PR #1757, opened 2022-11-01
remove mobile device check for nightMode stars since phones can run the new animations just fine

ysbrandB:
The old js was slow but css animations can also be done by mobile devices so I reenabled them for mobile devices :)


JonaMata (approved):
LGTM

PR #1762, opened 2022-11-02
fixed event-visibility in calendar and export function

ysbrandB:
Made the code more readable and fixed events with a publication time.

### DIFF
--- a/app/Http/Controllers/ExportController.php
+++ b/app/Http/Controllers/ExportController.php
@@ -48,8 +48,8 @@ class ExportController extends Controller
                 if ($user->can('board')) {
                     $data = Activity::all();
                 } else {
-                    $data = Activity::with('event')->get()->filter(function ($activity) use($user) {
-                        return $activity->event&& $activity->event->mayViewEvent($user);
+                    $data = Activity::with('event')->get()->filter(function ($activity) use ($user) {
+                        return $activity->event && $activity->event->mayViewEvent($user);
                     });
                     foreach ($data as $key => $val) {
                         unset($data[$key]->event);
@@ -73,7 +73,7 @@ class ExportController extends Controller
                 if ($user->can('sysadmin')) {
                     $data = Event::all();
                 } else {
-                    $data = Event::all()->filter(function ($event) use($user) {
+                    $data = Event::all()->filter(function ($event) use ($user) {
                         return $event->mayViewEvent($user);
                     });
                 }
--- a/app/Http/Controllers/ExportController.php
+++ b/app/Http/Controllers/ExportController.php
@@ -45,7 +45,7 @@ class ExportController extends Controller
                 $data = Achievement::all();
                 break;
             case 'activities':
-                if ($user->can('board')) {
+                if ($user->can('admin')) {
                     $data = Activity::all();
                 } else {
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/SearchController.php
```diff
@@ -74,7 +74,7 @@ public function search(Request $request)
             ['id', 'title']
         );
         foreach ($presearch_events as $event) {
-            if ($event->mayViewEvent(Auth::user())) {
+            if ($event->mayViewEvent(Auth::user()) || Auth::user()->can('board')) {
```
Jessseee, line 77: This shouldn't be necessary, right?

app/Http/Controllers/ExportController.php
```diff
@@ -45,11 +45,11 @@ public function export($table, $personal_key)
...
                 break;
             case 'activities':
-                if ($user->can('admin')) {
+                if ($user->can('board')) {
                     $data = Activity::all();
                 } else {
-                    $data = Activity::with('event')->get()->filter(function ($val) {
-                        return $val->event && $val->event->secret == 0;
+                    $data = Activity::with('event')->get()->filter(function ($activity) use ($user) {
+                        return $activity->event && $activity->event->mayViewEvent($user);
```
Jessseee, line 52: I am pretty sure you cannot make an activity without an event, right?

app/Models/Event.php
```diff
@@ -117,13 +117,9 @@ public function committee()
...
     {
-        //board may always view events
-        if($user && $user->can('board')){
-            return true;
-        }
-        //only show secret events if the user is participating
+        //only show secret events if the user is participating, helping or organising
         if($this->secret){
-            if($user && $this->activity && $this->activity->isParticipating($user)){
+            if($user && $this->activity && ($this->activity->isParticipating($user) || $this->activity->isHelping($user) || $this->activity->isOrganising($user))){
```
Jessseee, line 122: You could also add `$user->can('board')` to this check.

resources/views/event/display_includes/event_block.blade.php
```diff
@@ -1,4 +1,4 @@
-@if($event->mayViewEvent(Auth::user()))
+@if($event->mayViewEvent(Auth::user()) || Auth::user()->can('board'))
```
Jessseee, line 1: Same here, you can add this to mayViewEvent

app/Http/Controllers/ExportController.php
```diff
@@ -70,10 +70,12 @@ public function export($table, $personal_key)
                 $data = Company::all();
                 break;
             case 'events':
-                if ($user->can('admin')) {
+                if ($user->can('sysadmin')) {
```
Jessseee, line 73: This should be 'admin' as a user with the 'board' role may also view all events.

Jessseee (changes requested):

resources/views/event/display_includes/event_block.blade.php
ysbrandB: Nope, mayViewEvent is now also used for the calendar and you do want to give the board the possibility to not have all events in their calendar.

Jessseee (approved):
LGTM

PR #1765, opened 2022-11-06
Laravel Sail!

### LINKED ISSUE #1759: Laravel 8 is here so we can use laravel sail
It would be nicer to use this than our own docker etc. and Jonathan wanted to do it.

Jessseee:
Sorry @JonaMata , I didn't notice you were assigned to this issue when I started working on this, but please have a look at these changes. For the other @saproto/web-developers as well, please try and pull these changes and see if it works on your machine with the included instructions. If you run into any issues please contact me! Closes #1759.

> **⚠️ Attention!**
You might have permission issues when you have used `fix_permissions.sh`. Please return all rights back to your WSL2 user using `sudo chown -R $USER:$USER .` in the root directory of the project. This is of course no issue if you freshly cloned the repository.

> **Other note**
Laravel Sail currently only supports PHP 7.4, 8.0 and 8.1. The production server is currently on PHP 7.3, it is important that we test any compatibility issues that arise from this PHP version upgrade. However according to the [official documentation](https://www.php.net/manual/en/migration74.php) and my limited testing there should be no obvious issues. The same thing goes for Node which is locked to `16.x` which currently resolves to `16.18.0`, instead of `16.14.1`.


JonaMata:
> ⚠️ **ATTENTION!**
> The domain of the session cookie is now locked to the `COOKIE_DOMAIN` env variable. This variable has to be set BEFORE this branch is deployed.

JonaMata (approved):
LGTM

PR #1771, opened 2022-11-08
Fix search results not visible on iOS

### LINKED ISSUE #1739: User and product search field results don't show on IOS
Likely due to a weird CSS issue the result lists don't show.

JonaMata:
Fixes #1739


ysbrandB (approved):
LGTM

PR #1774, opened 2022-11-10
Fixed the ticket scanner not having an csrf token

ysbrandB:
Added CSRF token and fixed double import of assets,javascripts


Jessseee:
Fixed in #1775

PR #1777, opened 2022-11-12
Add a Withdrawal Seeder

Wilkuu-2:
**Adds a seeding mechanism for withdrawals in the OtherDataSeeder**
This lets the developers look for mistakes in styling, layout etc.,
without having to seed the withdrawals manually. 

*Note: This seeder does not seed for the local admin user*

### DIFF
--- a/database/seeders/OtherDataSeeder.php
+++ b/database/seeders/OtherDataSeeder.php
@@ -15,6 +15,7 @@ use Proto\Models\HashMapItem;
 use Proto\Models\Member;
 use Proto\Models\OrderLine;
 use Proto\Models\User;
+use Proto\Models\Withdrawal;
 
 class OtherDataSeeder extends Seeder
 {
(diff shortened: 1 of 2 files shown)


database/seeders/OtherDataSeeder.php
```diff
@@ -160,4 +173,73 @@ public function run()
         $seeder_end = microtime(true);
         echo "\e[32mSeeded:\e[0m    \e[1mOtherDataSeeder\e[0m (".round(($seeder_end - $seeder_start), 2).'s)'.PHP_EOL;
     }
+
+    // Withdrawal seeding helper method, based on store() method in WithdrawalController.php
+    // NOTE: Any updates in that method should be mirrorred in here
+    public function seedWithdrawals(int $index, int $maxMoney, int $maxOrderlinesInWithdrawal): int
```
Jessseee, line 179: I think it should be possible to just use the WithdrawalController::store() method if you create a request object. That would avoid us having to mirror any changes between this function and the controller. https://stackoverflow.com/questions/40262947/create-a-laravel-request-object-on-the-fly

Jessseee (changes requested):

Wilkuu-2:
This pull request is stale, so I will close it. 
This needs to be remade for the new version of Laravel anyway.

PR #1778, opened 2022-11-12
Made helper edit visible and blocked adding 1 committee as helper mul…

### LINKED ISSUE #1716: Can't delete helper committees once added to a event
Once a helper committee is added to an event. It is not visible when you edit the activity, which means you also can't delete the committee.

JonaMata:
…tiple times

Fixes #1716


Jessseee (approved):
LGTM

PR #1779, opened 2022-11-13
WIP: Fixed withdrawals for users without bank information

### LINKED ISSUE #1772: Withdrawal controller ignores users without bank information
Currently the withdrawal generation ignores users with no bank information.

The wanted behaviour would be to show the user that this person exists and to provide an option to add this persons debt to the next withdrawal or to accept the loss and set the users orderlines to paid (maybe a separate paid by association/loss field)

Jessseee:
Now showing orderlines of users without bank information (in case a whoopsy has happened) and added ability to mark a user's withdrawal as loss. Fixes #1772.

_Yes, I know the new DB column is misspelled but I prefer consistency._

### DIFF
--- a/app/Http/Controllers/WithdrawalController.php
+++ b/app/Http/Controllers/WithdrawalController.php
@@ -66,10 +66,8 @@ class WithdrawalController extends Controller
             if ($orderline->isPayed()) {
                 continue;
             }
+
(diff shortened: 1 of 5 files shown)


ysbrandB (approved):
LGTM

PR #1781, opened 2022-11-13
Fixed weird layout of markdown field toolbar

Jessseee:
and added some additional options to the toolbar.


ysbrandB:
image
why is the insert table button so big?

ysbrandB (approved):
LGTM

PR #1785, opened 2022-11-14
Feature/copy event with signup details

ysbrandB:
Adds the ability to make a copy of an event, and set the start date to a specified date. 
The copy automatically copies the event registration as well and moves all dates relative to the start date.
This means that if the signup date is a week in advance and you copy the event to a week later the new signup date will still be a week in advance relative to the new start date. Illuminates the need for a more complicated event templating system.


Jessseee (approved):
LGTM

PR #1786, opened 2022-11-15
Fix isalfredthere not displaying correct message when time is set

### LINKED ISSUE #1456: Isalfredthere displays wrong message
When Alfred should be there it displays that he is there instead.

JonaMata:
Fixes #1456


Jessseee (approved):
LGTM

Issue #1783, opened 2022-11-14
New OmNomCom Cookiemonster 


daniepyromanie:
Yoyo

I have a new design for the Cookiemonster in the OmNomCom screen :)) It serves as promo for the Alu vs. Christmaswrap drink on the 19th of December. Preferably, it would be nice to see it from the 28th of November onwards (until the 19th of December), but if that doesn't work for the whole duration that is fine ofcourse :)))
Let me know what works for u
Thanks in advance :D

k
![versie 1](https://user-images.githubusercontent.com/33750931/201674401-959fc42b-05cd-438f-91d0-2d460ef27a7b.png)
bye
xooxooxox
Dani

PR #1788, opened 2022-11-15
special koenkert for a special christmas occasion

ysbrandB:
added the christmaswrap koenkert to the omnomcom config.
implements #1783.
Will be removed again by me after the event has happened.


Jessseee (approved):
LGTM

PR #1789, opened 2022-11-15
removed the double import of redirect in the protube controller

ysbrandB:
fixes: "Symfony\Component\ErrorHandler\Error\FatalError
Cannot use Redirect as Redirect because the name is already in use"


Jessseee (approved):
Someone also checked out sentry, LGTM

PR #1791, opened 2022-11-16
fixed quote search pagination

ysbrandB:
title says it all, the get query params did not persist between the pages, now they do :)


Jessseee (approved):
LGTM

PR #1792, opened 2022-11-18
Feature/new theme

ysbrandB:
Implement the advent calender page and theme for the December Month.
Will be reverted after december


resources/views/website/layouts/assets/stylesheets.blade.php
```diff
@@ -1,5 +1,10 @@
-@if(Auth::check() && Auth::user()->theme)
+@if(Carbon::now()->month===Carbon::DECEMBER ||(config('app.env') != 'production'))
```
Jessseee, line 1: First off, this won't be activated on production. But please also add the option for users to turn this off.

Jessseee (changes requested):

resources/views/website/layouts/assets/stylesheets.blade.php
ysbrandB: We discussed this with Max and Jonathan but we did not want to make it a permanent theme, and we did not want to alter the database to make the choice of which theme you have (since if we remove it we will have to do a migration that everyone with theme 5 will be set back to theme 1 or another one), so we discussed it would be best to just force everyone over it for 1 month. Or do you have an easier way to do this?

MaxLiebe: I personally did not mind adding another theme, and just running the following query on the database when the holidays are over:
```UPDATE saproto.users u
SET u.theme=1
WHERE u.theme = 4```
Or whatever the id of the theme will be.
I still believe this is the best solution.

Comment (ysbrandB) [reply resources/views/website/layouts/assets/stylesheets.blade.php]: Added the ability to disable the theme by setting a cookie (button is found by the theme dashboard)

File: resources/assets/js/december.js [hunk 2]
```diff
@@ -0,0 +1,3 @@
+window.december = _ => {
+    console.log("DECEMBERMAAND!")
```
MaxLiebe, line 2: Uh

app/Http/Controllers/AdventController.php
```diff
@@ -0,0 +1,28 @@
...
+
+//        to test uncomment the two lines below!
+//        $ids=[2040,2,3, 1,5,2068, 7,2030,9];
+//        $date=Carbon::createFromFormat('Y-m-d H:i:s', '2022-10-19 12:05:00');
+
+        $events = Event::whereIn('id', $ids)->get();
+        return view('advent.index',['eventsArray'=>$events, 'date'=>$date]);
+    }
+
+    public function toggle_december() {
```
MaxLiebe, line 24: Style guides exist for a reason!

MaxLiebe (approved):
LGTM, but pay attention to styling & remove console logs.

Jessseee (approved):
LGTM

PR #1807, opened 2022-11-28
Enable/temp admin button

ysbrandB:
Protube works again with this functionality, so the button can be back


Jessseee:
Maybe don't do this from the Laravel 9 branch to master...

PR #1811, opened 2022-11-30
Implement shift-select feature and add it to the photo admin

JonaMata:


ysbrandB:
LGTM, would be nice to add that you can also extend the current selected photos with the arrow keys though

ysbrandB (approved):
LGTM

PR #1821, opened 2022-12-06
Feature/event attendees

### LINKED ISSUE #1812: add the possibility for the board to add to an event how many people actually showed up
It would be handy to keep track of this in the website and then combine it with the categories feature to make a nicer statistics page between two dates how many activities per event category between that time, percentage of expected people who showed up etc

ysbrandB:
Add the ability to estimate how many people actually attended an event. This makes manually keeping track of this in an overview redundant and is nice for statistics and ALV's.
Fixes #1812


MaxLiebe (approved):
LGTM, I think the decision to replace the helper column with the attendees column makes sense.

PR #1823, opened 2022-12-06
made password store use target="_blank"

### LINKED ISSUE #1798: Make links in password store use target="_blank"
It's mildly annoying that links in the password store do not open in a new tab at the moment

ysbrandB:
title says it all :)
Fixes #1798


Jessseee (approved):
LGTM

PR #1827, opened 2022-12-15
Updated birthday image of board 12.0

MaxLiebe:


ysbrandB (changes requested):
The board likes the old one better, please do not change

PR #1829, opened 2022-12-17
fixed the ability to set the photo album thumbnail

ysbrandB:
Title says it all, probably broke due to the shift select feature


Jessseee (approved):
LGTM

PR #1830, opened 2022-12-18
Feature/random public photo api endpoint

ysbrandB:
Made an endpoint where you fetch a JSON object with the URL to a random public photo, the date it is taken and the album name it belongs to.
Necessary to implement the feature on ProTube where we will show a random public photo when the queue is empty. See https://github.com/saproto/ProTube/pull/73 for the feature.


Menkveld-24 (approved):
LGTM

PR #1835, opened 2022-12-27
Fixed assets not loading in dev

ysbrandB:
fixed the .env.dev to redirect to localhost instead of our assets domain


Jessseee (approved):
LGTM

Issue #1658, opened 2022-09-09
Bulk update withdrawal page


BenLigthart:
Would be nice if I could select all the people who i.e. had a failed withdrawal so I do not have to do this for everyone individually. Something like a box I can check per name at the left side of the screen. Kusjes your penny x

PR #1836, opened 2023-01-02
Make the withdrawal overview have a bulk update for the fail and remove action

ysbrandB:
You can now select the users with a checkbox and them choose the action you want to perform.
new UI:
image
Implements #1658.


app/Http/Controllers/WithdrawalController.php
```diff
@@ -27,24 +27,17 @@
...
+    public function create(): View
     {
         return view('omnomcom.withdrawals.create');
     }
 
-    /**
-     * @return RedirectResponse
-     */
-    public function store(Request $request)
+    public function store(Request $request): RedirectResponse
```
JonaMata, line 40: Why are all the comments removed?

app/Http/Controllers/WithdrawalController.php
```diff
@@ -203,40 +181,78 @@ public function destroy(Request $request, $id)
     }
 
     /**
-     * @param  int  $id
-     * @param  int  $user_id
-     * @return RedirectResponse
+     * @param  int[]  $userIds
      */
-    public static function deleteFrom(Request $request, $id, $user_id)
+    public static function dSeleteFrom(Withdrawal $withdrawal, array $userIds): RedirectResponse
```
JonaMata, line 186: There's a typo in the name of this method.

resources/views/omnomcom/withdrawals/show.blade.php
```diff
@@ -130,73 +132,77 @@ class="btn btn-outline-success btn-block">
...
+                    @if(!$withdrawal->closed)
+                        <div class="card-footer bg-dark text-white mb-2 mr-2">
+                            <div class="d-inline-flex flex-row flex-row-reverse w-100">
+                                <button name="action" type="submit" value="remove" class="btn btn-danger">
+                                    Remove for selected users
+                                </button>
+
+                                <button name="action" type="submit" value="markfailed" class="btn btn-warning me-2">
+                                    Mark failed for selected users
+                                </button>
```
JonaMata, line 201: The single update buttons implement a confirm modal for this action. I think this should also be required for the bulk update action, as it is even more impactful.

resources/views/omnomcom/withdrawals/show.blade.php
```diff
@@ -130,73 +132,77 @@ class="btn btn-outline-success btn-block">
...
-
-                                            |
-
-                                            @include('components.modals.confirm-modal', [
-                                               'action' => route('omnomcom::withdrawal::markloss', ['id' => $withdrawal->id, 'user_id' => $data->user->id]),
-                                               'text' => 'Loss',
-                                               'title' => 'Confirm Marking Loss',
-                                               'message' => 'Are you sure you want to mark this withdrawal for '.$data->user->name.' as a loss? <b>This cannot easily be undone!</b>',
-                                               'classes' => 'text-white fw-bold underline-on-hover'
-                                            ])
```
JonaMata, line 191: Maybe still keep the single update buttons as this is easier to use when just removing or marking one user?

JonaMata (changes requested):
LAGTM

app/Http/Controllers/WithdrawalController.php
ysbrandB: Pint does this when you actually use the typed function return

resources/views/omnomcom/withdrawals/show.blade.php
ysbrandB: I removed them because when you do select orderlines and then remove one orderline your selection will be gone. I think only having one way is more clear.

resources/views/omnomcom/withdrawals/show.blade.php
ysbrandB: I removed them because when you do select orderlines and then remove one orderline your selection will be gone. I think only having one way is more clear.

ysbrandB: Yes I agree, I do not understand the confirm-modal enough yet to implement this though

resources/views/omnomcom/withdrawals/show.blade.php
ysbrandB: Yes I agree, I do not understand the confirm-modal enough yet to implement this though

ysbrandB:
The intended behaviour is very hard to implement with our current confirm modal. Therefore this will be postponed untill we have an actual front end

Issue #1817, opened 2022-12-01
Utwente account sync is broken


ysbrandB:
When you make a new account and try to sync your utwente account it does not seem to work.
Later addition: also add that you need to give us a non-utwente account so we have a way to contact them after they stop studying (#1815)

PR #1838, opened 2023-01-02
Make validation rule so users can not have an utwente email adress

### LINKED ISSUE #1815: Mandatory external e-mail (apart from UT mail) to prevent loss of contact
My dearest beheEeEeeEEeEr friends,

Another issue, another quote. This issue is about e-mail. So I will bless you once again with a personalized quote:

"Email has an ability many channels don't; creating valuable, personal touches - at scale" - David Newman

This wonderful quote can only confirm how important email is to us. Especially since I am no Tristan and calling is not my preference when contacting people, especially those who do not pay their Proto fees. This is what brings us to my issue! 

Every month, a couple of people do not pay their Proto withdrawal (I will not shame those people, but shame) due to many reasons. Some have been there for a very long while already. Today I wanted to contact these people, however a lot of members have their student mail connected to their account. Most of them are already done with being a student, which means their mail got deleted and I have no way of contacting them through mail. Therefore I was wondering whether it might be possible to add a mail check when people sign up as Proto member that they have at least one external e-mail address connected. I do not know if this clashes with connecting the Proto account to our UT account in any way, but just throwing it out there :)

Thank you in advance, this time I will not forget to order iced coffee for your next beheer meeting x

Kusjes,
Ur penny and personal quote corner

ysbrandB:
This is necessary because it happens too often that people stop studying here and our email rejects.
Note: this does not fix the issue when creating an account with your utwente account. When we fix that (#1817) we should look into making an extra screen where you input a non-utwente account as well.
Partially fixes #1815.

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -125,7 +125,7 @@ class AuthController extends Controller
 
         Session::flash('register_persist', $request->all());
         $this->validate($request, [
-            'email' => ['required','unique:users','email', new NotUtwenteEmail],
+            'email' => ['required', 'unique:users', 'email', new NotUtwenteEmail()],
             'name' => 'required|string',
             'calling_name' => 'required|string',
             'g-recaptcha-response' => 'required|recaptcha',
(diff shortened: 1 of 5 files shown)


JonaMata (approved):
LGTM

PR #1840, opened 2023-01-05
Update/Merge good idea board and quote corner and add extra features

ysbrandB:
Complete changed the goodidea board:

- Renamed everything to feedback as it now is used more widely
- Added feedback categories so every 'feedbackboard' has its own category
- Added the admin tools for the categories
- Added the ability for the board to reply to a good idea, accept or reject it and automatically send an email with a response
- Add the feature that the category can have a reviewer and not publish the feedback until it is approved
- Add a cron job and an email that reminds the reviewer if there are any unreviewed ideas.
- Add a command to migrate the QuoteCorner to the new feedback system completely (proto:migratequotes)
- let the old routes of /goodideas and /quotes redirect to /feedback/goodideas and /feedback/quotes

The removal of the quotesystem needs to be done in a seperate PR after the migration command is run
fixes #1252, fixes #1181, fixes #1232

PR #1841, opened 2023-01-05
Remove all quotecorner files and add to the gdpr export all feedback categories

ysbrandB:
Removed all files now unnecessary from the quote corner and fixed the GDPR export to include all feedback.
**First merge #1840 and run the command proto:migratequotes** before merging this into master.


MaxLiebe (changes requested):
- Some places have good idea hardcoded, which is odd.
- Removing e.g. a quote asks whether you want to delete this "potentially good feedback", which while being funny, is not really relevant
- The migratequotes command does nothing if a category named quotes already existed, which is because of the following:
- Duplicate feedback categories can exist, which can break things
- Make sure to not seed quotes anymore in the seeder
- A feedback category for quotes should be created by default to not make the menu item lead to a 404 page :)

Jessseee:
PuLL ReqUeST iS OutDaTed!1!!

ysbrandB:
Became outdated with #1971

PR #1845, opened 2023-01-07
Add from and to query to orders api and add image_url to product

JonaMata:


app/Http/Controllers/UserApiController.php
```diff
@@ -62,9 +63,21 @@ public function getAchievements()
     /**
      * @return Collection|OrderLine[]
      */
-    public function getPurchases()
+    public function getPurchases(Request $request)
```
Jessseee, line 66: Does this not break any other uses of this method?

Jessseee (changes requested):

app/Http/Controllers/UserApiController.php
JonaMata: Are there other usages?

Jessseee: Yeah, not sure, just checking.

Jessseee (approved):
LGTM

PR #1851, opened 2023-01-11
Update api naming

Menkveld-24:
Update for the api for ProTube database restructure pr, see https://github.com/saproto/ProTube/pull/52
Only merge them together as this will be breaking for non database refactored protube


ysbrandB (approved):
LGTM

PR #1856, opened 2023-01-13
Exclude the image url from the products in the live data seeder

ysbrandB:
The appends 'image_url' from the omnomcom wrapped broke the livedataseeder. This fixes it.


Jessseee:
Lol I pushed this to another branch but forgot to make a PR

Jessseee (approved):
LGTM

PR #1857, opened 2023-01-14
Actually implement the getAttendees method referenced in the earlier …

ysbrandB:
Actually implement the getAttendees method referenced in the earlier feature request of Feature/event_attendees that I totally not forgot to commit last time :).


MaxLiebe (approved):
lgtm

PR #1859, opened 2023-01-20
Member sign-out with delay

### LINKED ISSUE #1536: Member sign-out with delay
If I want to sign-out a member at this moment, I only have the option to sign them out immediately. It would be nice if there was an option to sign them out by the 31st of September (or somewhere around there before the new membership-fee withdrawal). They would (probably) need to get a new status on the site (member till ...), and it would be nice if they would get an automatic email with the fact that they will be signed out automatically before the start of the new college year.

ysbrandB:
It adds the ability to stop someone's membership at an arbitrary date and adds an easy button to end it at the end of September, which is right before October when we renew our memberships and charge the fees.
Fixes #1536.

### DIFF
--- a/app/Http/Controllers/UserAdminController.php
+++ b/app/Http/Controllers/UserAdminController.php
@@ -11,7 +11,7 @@ use Illuminate\View\View;
 use Mail;
 use PDF;
 use Proto\Mail\MembershipEnded;
-use Proto\Mail\MembershipStarted;
+use Proto\Mail\MembershipEndSet;
 use Proto\Models\HashMapItem;
 use Proto\Models\Member;
 use Proto\Models\User;
@@ -25,7 +25,7 @@ class UserAdminController extends Controller
      * @param Request $request
      * @return View
      */
-    public function index(Request $request)
+    public function index(Request $request): View
     {
         $search = $request->input('query');
         $filter = $request->input('filter');
@@ -198,7 +198,7 @@ class UserAdminController extends Controller
         $member->proto_username = $alias;
         $member->save();
 
-        Mail::to($user)->queue((new MembershipStarted($user))->onQueue('high'));
+        Mail::to($user)->queue((new MembershipEndSet($user))->onQueue('high'));
 
         EmailListController::autoSubscribeToLists('autoSubscribeMember', $user);
 
@@ -223,7 +223,7 @@ class UserAdminController extends Controller
      * @return RedirectResponse
      * @throws Exception
      */
-    public function endMembership($id)
+    public function endMembership($id): RedirectResponse
     {
         /** @var User $user */
         $user = User::findOrFail($id);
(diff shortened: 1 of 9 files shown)


app/Http/Controllers/UserAdminController.php
```diff
@@ -198,7 +198,7 @@ public function addMembership($id, Request $request)
         $member->proto_username = $alias;
         $member->save();
 
-        Mail::to($user)->queue((new MembershipStarted($user))->onQueue('high'));
+        Mail::to($user)->queue((new MembershipEndSet($user))->onQueue('high'));
```
Jessseee, line 201: I don't think we want to send a MembershipEndSet email whenever a membership is created.

database/migrations/2023_01_20_211900_add_member_until_to_member.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('members', function (Blueprint $table) {
+            $table->bigInteger('member_until')->after('is_pending')->nullable();
```
Jessseee, line 17: Naming the column `member_until` in the `members` table seems redundant, maybe name the column `until` instead.

resources/views/users/admin/admin_includes/membership.blade.php
```diff
@@ -26,6 +34,16 @@ class="list-group-item">
                     Preview membership card
                 </a>
 
+                @else
+                    @include('website.layouts.macros.confirm-modal', [
+                           'action' => route("user::member::removeend", ['id'=>$user->id]),
+                           'method' => 'POST',
+                           'classes' => 'list-group-item text-danger',
+                           'text' => 'Remove membership end!',
+                           'message' => "Are you sure you want to remove then end of the membership of $user->name?"
```
Jessseee, line 43: This is not a correct sentence.

resources/views/users/admin/admin_includes/membership.blade.php
```diff
@@ -26,6 +34,16 @@ class="list-group-item">
                     Preview membership card
                 </a>
 
+                @else
+                    @include('website.layouts.macros.confirm-modal', [
+                           'action' => route("user::member::removeend", ['id'=>$user->id]),
+                           'method' => 'POST',
+                           'classes' => 'list-group-item text-danger',
+                           'text' => 'Remove membership end!',
```
Jessseee, line 42: I think this is a confusing button text. Maybe "Cancel membership removal" would be a better phrasing.

resources/views/emails/membershipenddate.blade.php
```diff
@@ -0,0 +1,29 @@
...
+<p>
+    Dear {{ $user->name }},
+</p>
+
+<p>
+    I am writing to you to let you know that the secretary has set an end date for your membership at S.A. Proto.
+</p>
+<p>
+    Your membership will end at: <b>{{Carbon::createFromTimestamp($user->member->member_untill)->format('d-m-Y')}}.</b><br>
+    This means you are a full member before this date, but will not be for the following academic year.
```
Jessseee, line 14: I think this should be phrased differently as the end date might be different than September so saying that the membership is no longer valid in the following academic year might be confusing. Maybe change it to: "This means that your membership is valid up until this date and will then be terminated."

Jessseee (changes requested):

Jessseee (approved):
LGTM

PR #1861, opened 2023-01-22
Add a board email to the member untill functionality and fix the cronjob

ysbrandB:
The cronjob accidentally checked a carbon object against an integer.
Also added an email that updates the board with which memberships have been terminated.

image


Jessseee (approved):
LGTM

PR #1871, opened 2023-02-07
fix the ENV check for the fishcam

ysbrandB:


Jessseee:
Should we still use `fopen` to read the stream or could we just use the HTTP facade?

Lupus1 (approved):
Ziet er nice uit, mooie grote comit

Frank-Bosman (approved):
Poging twee met me goede account

PR #1876, opened 2023-02-08
Fix activity relation breaking ImportLiveDataSeeder for non-admin developers

Wilkuu-2:
This seems to fix the error:
```
Object of class stdClass could not be converted to string
```
This error occurs when inserting imported events. 

I don't see the reason why this only occurs on non-admin accounts, but this seems to fix it for them.


Wilkuu-2:
De `activity` relation wordt nu aan beide kanten verwijderd. 
Als dat een probleem is kan ik ook de eerste commit reverten.

ysbrandB (approved):
LGTM

PR #1877, opened 2023-02-08
Update opening times in emails

Wilkuu-2:
It now looks approximately like this:
image

I'm not sure where the official time can be referenced, so it is taken from the website.
(Mon and Fri -> 8:30-16:00
 Tue to Thu -> 8:30-17:30)


ysbrandB (approved):
LGTM

Issue #1868, opened 2023-01-27
Calendar shows participating also when on backup-list


ysbrandB:
When you sign up for an event and are placed on the backup list, your calendar will show that you are participating which is confusing since you do not have a spot. I would suggest replacing the [participating] with [backup-list] or something to make it clear that you do not have a place.
![image](https://user-images.githubusercontent.com/72889753/215055707-cacbfc7b-1af4-4e84-936c-d766ad68e12e.png)

PR #1878, opened 2023-02-09
Add back-up list check for setting event status

Wilkuu-2:
This branch makes it, so the exported calendar properly denotes if the user is on the back-up list for an event. Fixes issue #1868.

### DIFF
--- a/app/Http/Controllers/EventController.php
+++ b/app/Http/Controllers/EventController.php
@@ -560,7 +560,7 @@ class EventController extends Controller
                     if ($event->activity->isHelping($user)) {
                         $status = 'Helping';
                         $info_text .= ' You are helping with this activity.';
-                    } elseif ($event->activity->isOnBackupList()){
+                    } elseif ($event->activity->isOnBackupList($user)){
                         $status = 'On back-up list';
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/EventController.php
```diff
@@ -560,7 +560,10 @@ public function icalCalendar($personal_key = null)
                     if ($event->activity->isHelping($user)) {
                         $status = 'Helping';
                         $info_text .= ' You are helping with this activity.';
-                    } elseif ($event->activity->isParticipating($user) || $event->hasBoughtTickets($user)) {
+                    } elseif ($event->activity->isOnBackupList()){
```
ysbrandB, line 563: The isOnBackupList function takes a user as an argument, so this will not work

ysbrandB (changes requested):

app/Http/Controllers/EventController.php
Wilkuu-2: PHPStorm did a little trolling :rage::rage::rage: 

Commit 366adf3 should fix this.

ysbrandB (approved):
LGTM

Issue #1867, opened 2023-01-26
Wall street drink


Sebastiaanjai:
A wall street drink! 

![image](https://user-images.githubusercontent.com/48732717/214863701-e9db9484-f2ce-46fd-bb6d-6ac7496a7070.png)

Basically, a drink where the prices are determined by how much each product is bought, like stocks.

So, some starting things:
Each product should have its own separate price, which will decrease over time and will go back up when someone buys a specific drink. So somehow our purchase system should directly influence the price of products in the omnomcom.

Since every product might be too much I have selected a few that would be interesting for the drink:
- TIPcie Grolsch Premium (Glass) 
- TIPcie Weizen (big and small, but for this they can count as the same product. The price of Weizen big can just be the price of a small weizen x 1.35
- TIPcie Shots (normally nozem would be different from other shots but it is less effort to make them the same product)
- TIPcie Soda (big and small can be the same price)
- TIPcie Wine glass / bottle ( a bottle is 4.8x the price of a glass)
- TIPcie Apple Bandit
- SBZ Monthly SpecialBeer (unknown yet for that time, but ill notify you)

Each product can have a starting price equivalent to its current price, but can decrease in price each minute by 2 cents.
When the product is purchased the price should increase by 5 cents (per purchase, so if someone orders 10 beers the price should just go up by 5 cents). 
Each product should never exceed its starting price.
Each product should also never cost less than 10 cents.

We should do a bit of playtesting to see if these prices are accurate.

Additional idea but not strictly necessary:
A scrolling LED screen that displays the current prices of items

![image](https://user-images.githubusercontent.com/48732717/214869534-d6c7edc0-806f-49de-9164-848477b648fb.png)
something like this. But a bit smaller.

Inter-Actief has done something like this in the past. You can see their project here for inspiration:
https://github.com/inter-actief/alexia/tree/wallstreet 

The drink itself is now planned for the 3rd of april.

Good luck! If you need anything else from me let me know.

PR #1879, opened 2023-02-11
Wallstreet Drinks!

ysbrandB:
Implements #1867.

Adds the functionality of Wallstreet Drinks.
Made way more modular than suggested.

In the Wallstreet drink overview you can create new drinks and add the items whose price should fluctuate to the new drinks,
this is also where you set the minumum price and price increase and decrease for that specific drink.

The Cronjob only runs when a drink is active and updates the price every minute.
In the admin panel also a chart is implemeted under the "view price history" button that shows a line chart for that drink per item with the price over time. 

An endpoint (api/wallstreet/updated_prices) is made which gives you the latest prices when a drink is active and how much this item has gone up or down, which is also used for the updating of the TIPCie screen so they know what price the products are.

Also a screen was implemented using this endpoint to show the products and prices of the ongoing drink. (see below)

The admin overview and price chart:
image
image

The wallstreet marquee sign: (note: the background of the divs is the product picture but locally I do not have those)
image


Frank-Bosman:
First of let's start of by saying: it looks great!!
But there are a few small problems:
 - The prices on the omnomcom and marquee screen somethimes randomly go back to their default value eventhough they are correctly displayed in the charts. It is often just one or two products who do this.
 - The input when making a drink also isn't checked so you can enter a start time which is after your end time.
 
 There are also a few styling things:
 - The tab index is skuffed and messed up for the edit/create form. after you click tab on the start date you should go to the end date
 - The `WallstreetDrink overview` title in the overview screen isn't readible in light mode

Frank-Bosman (approved):
LGTM, We will probably find more bugs after the test session. But it's good for now

Issue #1385, opened 2021-10-07
Custom omnomcom sounds


PetervanderBurgt:
I mean, how great would it be if members could have their own custom sound, instead of good ol cookie monster munchin away?

Looking forward to your reaction,

Peter

PR #1880, opened 2023-02-12
Custom omnomcom sounds!

ysbrandB:
Implements #1385.
If you set a new custom sound (mp3, no larger than 200kb) in your dashboard, it will mute the cookiemonster video and play that sound instead!

### DIFF
--- a/app/Http/Controllers/OmNomController.php
+++ b/app/Http/Controllers/OmNomController.php
@@ -163,6 +163,10 @@ class OmNomController extends Controller
             return json_encode($result);
         }
 
+        if($user->member->customOmnomcomSound){
+            $result->sound = $user->member->customOmnomcomSound->generatePath();
+        }
+
         if ($user->disable_omnomcom) {
             $result->message = "You've disabled the OmNomCom for yourself. Contact the board to enable it again.";
             return json_encode($result);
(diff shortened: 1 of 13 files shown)


Jessseee:
I just want to say that I do not agree with this feature as it has huge potential for abuse and does not add anything of value, in my opinion.

ysbrandB:
> I just want to say that I do not agree with this feature as it has huge potential for abuse and does not add anything of value, in my opinion

Would you feel differently when the board has to set your custom sound and therefore the abuse part of it is negated?

Jessseee:
> > I just want to say that I do not agree with this feature as it has huge potential for abuse and does not add anything of value, in my opinion
> 
> Would you feel differently when the board has to set your custom sound and therefore the abuse part of it is negated?
> I think then the customization does add to the site just like we do with the custom welcome messages.

Yes, I think that would be fine.

ysbrandB:
> > > I just want to say that I do not agree with this feature as it has huge potential for abuse and does not add anything of value, in my opinion
> > 
> > 
> > Would you feel differently when the board has to set your custom sound and therefore the abuse part of it is negated?
> > I think then the customization does add to the site just like we do with the custom welcome messages.
> 
> Yes, I think that would be fine.

Refactored so the board uploads a sound in the user admin instead for a member :)

Frank-Bosman (approved):
Ziet er goed uit!!

PR #1881, opened 2023-02-12
Heigthen the value of max_input_vars for the handle function of DA-sync

ysbrandB:
Hopefully fixes the sentry error that the DA sync goes over the parse_string limit of 1000.


Jessseee (approved):
LGTM

Issue #1869, opened 2023-01-31
Clarify outstanding amount


ysbrandB:
The outstanding amount is a usefull number but as of right now not clear enough.
Also it would be nice if in a dropdown or tooltip it would show in which activities the money is and how much per activity.

ysbrandB:
Fixed in #1885

ysbrandB:
Fixed in #1885

PR #1885, opened 2023-02-18
Clarify the outstanding amount and make it so you can see the activies which still have to be processed

ysbrandB:
Implements #1869 
The clarification under the tooltip:
image
The activity accordion when clicked on the amount:
image


Menkveld-24 (approved):
LGTM

PR #1888, opened 2023-02-27
Laravel 10

ysbrandB:
On the feature branch for Laravel 9 the update to Laravel 10 was quite straightforward.
Followed the upgrade guide and updated packages.
Renamed routeMiddleware as suggested and removed the deprecated $dates for $casts[].
Removed registerPolicies() and Passport::routes() as they are no longer required.
The Hashid packaged now seems to need an int as the length argument so updated that as well.


ysbrandB:
@Jessseee I believe I have fixed and updated everything now to a stable working version. Could you have a look again through this (and therefore the Laravel 9 branch which I also updated) when you have the time?

Jessseee (approved):
LGTM

PR #1890, opened 2023-03-01
Fix the copying of events

ysbrandB:
When an event with a closed activity was copied it would simply copy that status.
Now it creates a new activity and only copies the necessary info.
Also made it so the copy includes the helping committee instances and fixed the redirect to actually go to /event/edit
instead of /event/copy?id which made the subsequent requests fail (like editing the new event etc).


MaxLiebe (approved):
LGTM, but redirecting is kinda iffy

PR #1891, opened 2023-03-01
Tempadmin for ProTube

Menkveld-24:
Api update for https://github.com/saproto/ProTube/pull/114


app/Models/User.php
```diff
@@ -596,10 +596,16 @@ public function getSignedMembershipFormAttribute()
         return $this->member && $this->member->membershipForm !== null;
     }
 
-    /** @return bool */
-    public function getIsProtubeAdminAttribute()
```
Menkveld-24, line 600: This function was unused to I repurposed it

PR #1892, opened 2023-03-03
Move over the web route files to the new syntax (laravel 8)

ysbrandB:
Moved the syntax from web.php over to the new php callable syntax and made more appropriate use of grouping.
I suggest taking a work evening to test this thoroughly.

For some reason 'sail artisan route:list' can not find the ProtubeController and I do not see why as the filenames seem to be correct and the import and namespace are there and my editor recognizes it. So if someone sees my mistake please let me know.
Also reminded me of the fact that we need to remove all of the unused Protube features as more than half the routes are old and unused.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -49,6 +49,7 @@ use Proto\Http\Controllers\RegistrationHelperController;
 use Proto\Http\Controllers\RfidCardController;
 use Proto\Http\Controllers\SearchController;
 use Proto\Http\Controllers\ShortUrlController;
+use Proto\Http\Controllers\SmartXpScreenController;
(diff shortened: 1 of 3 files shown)


routes/web.php
```diff
@@ -10,853 +72,944 @@
     View::share('viewName', $view->getName());
 });
 
-// TODO: change string based controller route definitions to standard PHP callable syntax.
-
-Route::group(['middleware' => ['forcedomain']], function () {
+ROute::middleware('forcedomain')->group(function () {
```
JonaMata, line 75: THis HAs A CApital LEtter TOo MAny

ysbrandB:
wayy to outdated atm

PR #1894, opened 2023-03-07
set the uploadlimit for our local phpmyadmin image so it does not default to 0

ysbrandB:
When trying to import a database locally for debugging the phpmyadmin errors because no upload limit is set in the docker container so it defaults to 0 bytes. This is not taken from the php.ini because it is its seperate container. Now set to 500mb so we will not encounter the limit again soon.


MaxLiebe (approved):
Bruh

PR #1895, opened 2023-03-09
Fix/improve committee page visibilty

Frank-Bosman:
## Improved committee, photo and societies pages

Splitted the overlay on the events to a new css class, `.dark-card`. This class is then used on the committee, societies and photo overview pages, as well as the events in on the home page.
Also updated the default white overlay to have a lower opacity so the page isn't completly white.
The rainbowbarf has a custom beatifull animated overlay.

afbeelding


ysbrandB (approved):
LGTM (Rachel is heel trots op je!)

PR #1897, opened 2023-03-13
Decreased overlay intensity

Frank-Bosman:
## Decreased overlay intensity
The overlay was way to intense on the dark mode so the opacity has been lowered.
Please look if it is now good on the commitee page and in the photo albums.


ysbrandB:
This is based upon the wrong way of implementing the photo pop, now fixed. If you want to have a differen amount for dark and light then add an extra class.

PR #1911, opened 2023-04-01
Updates to wallstreet marquee screen!

ysbrandB:
Updates to the marquee screen. Now includes the graph and in general is just better designed.
Also per TiPcIE request the cap of the wallstreet price has gone to 120% of the original price.
The marquee screen now looks like this:
image


Wilkuu-2 (approved):
LGTM

Issue #1820, opened 2022-12-06
Two separate sign-up lists for activities


Sebastiaanjai:
For a very specific event the TIPcie would like to have an event with two sign-up lists with separate names. So not two tickets, the idea is that the activity is free but we would like people to choose a side and publicly list who chooses what side.

This activity miiiight be on the 19th already but i understand that thats very fast. But this will be usefull again next year and possibly for other activities.

Sebastiaanjai:
I would assign this low-prio if i could

PR #1915, opened 2023-04-23
Add the option to show the users who have bought tickets on an event

### LINKED ISSUE #1900: Option to not add a signup but to show all members who bought a ticket
RIght now to do this we do and tickets and a signup which a lot of users find confusing.
I think it would be nicer to add the option to show the people who bought a ticket as participants in events

ysbrandB:
This helps with the confusion when for instance with a symposium you want to create hype but users get confused by only getting a ticket or only signing up. The idea is that you then will not make an event for that anymore. Fixes #1900 and #1820 as you can add two free tickets.


MaxLiebe (approved):
lgtm

PR #1916, opened 2023-04-24
Add soccercards message to OmnomCom

### LINKED ISSUE #1914: Soccercards notification in the omnomcom
What page or functionality is your issue about?
- the message in the omnomcom which states how many soccer cards you may get
What were you trying to do?
- now we want a cap on the maximum amount of cards so max 12 cards (0.50 cents per card so max 6 euros) 
- so if someone buys for 8 euro it is still 12 cards
What did you expect would happen?
- now there is no maximum
What did happen instead?
- because it is not active now nothing happend but it would have said 16 card for the 8 euro example
Where there any error or confirmation messages?
- no
Screenshots can provide a lot of extra information!
- i can not provide a screenshot of the omnomcom and also because it is not active yet/anymore

RubenKoole:
Fixes #1914


ysbrandB (approved):
LGTM

PR #1924, opened 2023-05-02
Fixed the user count on events

ysbrandB:
The user count accidentally also included helpers (which was not apparent because sometimes helpers are signed up and sometimes not and we only count the unique ID's).
Also, use the new checkbox component for the corresponding ticket toolie.


MaxLiebe (approved):
lgtm

PR #1928, opened 2023-05-14
Add the ability to fill out which user ordered to the dinnerform

### LINKED ISSUE #1904: Improvements on dinnerform
Add an option to fill in who ordered the dinner form. Because sometimes this is not clear anymore after a few weeks.

When you click on a dinnerform and open it on a new page, the title is missing.

ysbrandB:
Fixes #1904


app/Models/Dinnerform.php
```diff
@@ -61,6 +62,12 @@ public function event()
         return $this->belongsTo('Proto\Models\Event');
     }
 
+    /** @return BelongsTo */
+    public function orderedBy()
+    {
+        return $this->belongsTo('Proto\Models\User', 'ordered_by');
+    }
```
JonaMata, line 69: both this method and the database column result in the ordered_by property. It would be best to have different names for the database column and this method. For example a user_id column or ordered_by_user_id column name would work.

database/migrations/2023_05_08_143425_add_userid_to_dinnerform.php
```diff
@@ -0,0 +1,27 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('dinnerforms', function (Blueprint $table) {
+            $table->integer('ordered_by')->nullable();
```
JonaMata, line 14: Why is this nullable?

resources/views/dinnerform/admin_includes/dinnerform-list.blade.php
```diff
@@ -52,6 +53,16 @@
                         <td>{{ $dinnerform->start->format('Y m-d H:i') }}</td>
                         <td>{{ $dinnerform->end->format('Y m-d H:i') }}</td>
                         <td>€{{ number_format($dinnerform->totalAmountWithDiscount(), 2) }}</td>
+                        @if($dinnerform->ordered_by!=null)
```
JonaMata, line 56: By removing the whole <td> element the items are shifted to the left when there is no ordered_by value. Change this to an empty <td> instead.

JonaMata (changes requested):
LAGTM

database/migrations/2023_05_08_143425_add_userid_to_dinnerform.php
ysbrandB: It is nullable so all the existing dinnerforms do not default to user 0

JonaMata (approved):
LGTM

PR #1929, opened 2023-05-15
Prevent the tickets from being added to an withdrawal when they are still purchasable

### LINKED ISSUE #1903: Add tickets to a withdrawal only after the signup deadline has surpassed
When there are tickets being sold the Treasurer can accidentally make a withdrawal and put all the tickets on "paid" without actually intending it. This means that the tickets are not editable anymore even by the board. I think the best fix is to only let tickets be able to be added to the withdrawal when the signup has ended, or make a button so the treasurer actually has to decide to add the tickets. The first option is already discussed during our meeting and with the treasurer and I think is the easiest one.

ysbrandB:
Fixes #1903.


Wilkuu-2 (approved):
LGTM!

PR #1931, opened 2023-05-15
Start tracking manual stock modifications

### LINKED ISSUE #1852: OmNomCom changes marked as 'loss'
Lovely beheer friends,

First of all, happy new year! I am happy to be able to write my first issue of 2023. New year new me, but writing issues will remain the same as usual :). This issue will be about loss. Therefore, I will give you an inspiring quote about loss:

"Things we lose have a way of coming back to us in the end, if not always in the way we expect" - J.K. Rowling

In my case, this means it comes back in sad bookkeeping noises. Whenever an item is expired or a box is completely missing, I have to remove this from the OmNomCom system so people cannot buy it anymore. However, when I remove these items it means that I can not mark them as loss at the module count of the OmNomCom (at the start of every module we do a big count how many items were stolen/forgot to be striped). If I would want to maintain this the treasurer would have to keep a long list of everything they ever changed in the OmNomCom, which is a lot of work and impossible to do since other OmNomCom members can change the stock as well. 

Therefore I was wondering whether an automatic list could be generated with these items. So for example I want to remove 30 pink cakes I can do so and click on a box which is called "Mark as loss", which will add the product change to a list of lost items. 

Again, thank you so much for your time and effort <3 see you!

Kusjes,
Ben

Wilkuu-2:
This PR makes it so ALL changes in the Omnomcom stock are tracked. 

It also allows the user to view, filter and download reports on manual changes. 

Fixes #1852


Wilkuu-2:
Feedback: 
* Clarify the new stock edit menu changes
* Make the "Show only losses" filter the default
* Change the id in csv to the product name

Bugs: 
* Check why the order of mutations is reverse in a single edit

Additional bug: 
* Double checkbox in product edit

ysbrandB (approved):
LGTM

PR #1932, opened 2023-05-15
Update the fishcam so it only hits the proxy directly and not the server anymore

ysbrandB:
On merge the FISHCAM_URL in the .env should be changed to "https://utils.saproto.nl/fishcam"


Wilkuu-2 (approved):
LGTM!

PR #1933, opened 2023-05-16
Use the FISHCAM_URL instead of the removed API route in Smartxp Screen

Wilkuu-2:
Fixes the screens in SmartXP. 

They broke because the fishcam API is no longer there.

This PR fixes it. 

Breaking PR: #1932


Frank-Bosman (approved):
LGTM

PR #1936, opened 2023-05-23
Update the php version of the .env files to 8.2

ysbrandB:
We are running this locally and on the server.
This will make our deploy check and deploy actually work again as well, as they use this to setup PHP.
Before:
image
Now:
image


JonaMata (approved):
LGTM

PR #1937, opened 2023-05-24
Fixed the reference to the selectable photo view

ysbrandB:
This fixes the photo-upload.


app/Http/Controllers/PhotoAdminController.php
```diff
@@ -106,7 +106,7 @@ public function upload(Request $request, $id)
             $uploadFile = $request->file('file');
 
             $photo = $this->createPhotoFromUpload($uploadFile, $id);
-            return html_entity_decode(view('website.layouts.macros.selectablephoto', ['photo' => $photo]));
+            return html_entity_decode(view('photos.includes.selectablephoto', ['photo' => $photo]));
```
JonaMata, line 109: What about moving the selectable photo macro to the macros folder?

JonaMata (changes requested):
LAGTM

app/Http/Controllers/PhotoAdminController.php
ysbrandB: Jesse has moved it here in the Laravel 9 update. The macros folder does not exist anymore

Menkveld-24 (approved):

PR #1938, opened 2023-05-30
Feature/Implement a redirect after you sign up

### LINKED ISSUE #1813: Forced to click on link when signing up for events.
Dear HYTTIOAOAC,

For external events sometimes links with a form made by the company is required to be filled in.
So I would like to have a some sort of pop-up or place where people are forced to click on the link or be reminded to click on the link when signing up for an external event.

ysbrandB:
This implements the functionality that you can specify a redirect URL when making an event. When the user signs up they get redirected here so they can, for instance, fill out a form.
Fixes #1813

### DIFF
--- a/app/Http/Controllers/ParticipationController.php
+++ b/app/Http/Controllers/ParticipationController.php
@@ -73,8 +73,7 @@ class ParticipationController extends Controller
         $participation->fill($data);
         $participation->save();
 
-
-        if(!$is_web){
+        if (! $is_web) {
             if ($event->activity->isFull() || ! $event->activity->canSubscribe()) {
                 $message = 'You have been placed on the back-up list for '.$event->title.'.';
             } else {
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/ParticipationController.php
```diff
@@ -72,20 +73,25 @@ public function create($id, Request $request)
...
-        } else {
+        if (! $is_web) {
             if ($event->activity->isFull() || ! $event->activity->canSubscribe()) {
                 $message = 'You have been placed on the back-up list for '.$event->title.'.';
             } else {
                 $message = 'You claimed a spot for '.$event->title.'.';
             }
-            abort(200, json_encode((object) [
+
+            return Response::JSON([
```
Menkveld-24, line 83: Isn't it just Response::json() or response()->json()? Without using the "use Symfony\Component\HttpFoundation\Response;" but "use Response;" instead?

resources/views/event/edit_includes/activity.blade.php
```diff
@@ -81,6 +81,18 @@
                                    required>
                         </div>
 
+                        <!-- Website -->
```
Menkveld-24, line 84:
```suggestion
```

Menkveld-24 (approved):

Menkveld-24 (approved):
LGTM

PR #1939, opened 2023-06-02
Made the endpoint

ysbrandB:
Temporary endpoint to show photos older than 4 years for the alumni event


MaxLiebe (approved):
lgtm (function is temporary)

PR #1942, opened 2023-06-19
Fix the narrowcasting adding infinite divs and the ability to add youtube videos

ysbrandB:
Removed the YT library from the narrowcasting to add back the ability to add youtube videos.
It also fixed the hiding and unhiding of the slides and prevented the slides from being added infinitely.
Also added that when there is a slide added or unadded it reloads so it always has the correct data.


ysbrandB:
All included in #1946 already

PR #1946, opened 2023-06-23
Feature/new protopolis screen

ysbrandB:
A new screen for in the Protopolis on the left TV screen. This combines the Protopolis info and the narrowcasting, as we no longer have separate screens for them.
Adds the screen and implements the necessary code for it.
It also fixes the narrowcasting.

The screen looks like this (A too long activity title now truncates not scrolls but that is not yet reflected in the screenshot):
protopolisScreen


resources/views/smartxp/protopolis_screen.blade.php
```diff
@@ -0,0 +1,328 @@
...
+    </style>
+
+    </style>
+@endpush
+
+@section('body')
+    <div class="row text-black">
+
+        {{-- narrowcasting--}}
+        <div class="col-9" style="border-radius: 8px; padding:0; overflow:hidden;  border: 0px solid #00AAC0;
```
Jessseee, line 65: Why all this inline css?

resources/views/smartxp/protopolis_screen.blade.php
```diff
@@ -0,0 +1,328 @@
...
+                              newDiv.style.backgroundRepeat = 'no-repeat'
+                          }
+
+
+                          let titleDiv=document.createElement("div")
+                          titleDiv.className="overflow-hidden"
+
+                          let titleSpan=document.createElement("span")
+                          titleSpan.innerHTML=activity.title
+                          titleSpan.className="text-nowrap text-truncate w-100 me-5 font-weight-bold font-size-lg"
```
Jessseee, line 229: I think a scrolling title would be preferable. See the ProTube screen for an example.

Jessseee (changes requested):
LAGTM, nice design btw.

resources/views/smartxp/protopolis_screen.blade.php
ysbrandB: Implemented :)

Jessseee (approved):
LGTM

PR #1947, opened 2023-06-24
Fix the first element never scrolling and make the narrowcasting 100vw and vh

ysbrandB:
Due to the first activity flexing to the full width and then doing the overflow check it would never scroll.
Now populate the activities first and then do the checks.
Also add the forgotten CSS of 100vw and 100vh to the narrowcasting


MaxLiebe (approved):
Looks GT Me

PR #1950, opened 2023-06-25
Fix people receiving helper reminders when not member of the committee

### LINKED ISSUE #1608: Helper emails dont stop
Hoi haveyoutried,

the helper emails (for protography) keep on being sent to me while Im no longer an active member.
Maybe, keeping proto's sender score in mind, it is a good idea to make sure this does not happen so your emails are not marked as spam (as often).

Groetjes thuis!

### LINKED ISSUE #1689: Automatic unsubscribe helper reminder mails
When people are subscribed to helper reminders and they stop being part of a committee, they can't unsubscribe from the helper reminders anymore. Let's fix that (and fix current problems as well)

JonaMata:
Fixes #1689. Fixes #1608

### DIFF
--- a/app/Http/Controllers/CommitteeController.php
+++ b/app/Http/Controllers/CommitteeController.php
@@ -59,9 +59,7 @@ class CommitteeController extends Controller
             abort(404);
         }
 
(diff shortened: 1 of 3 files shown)


emielsteegh:
😍

PR #1970, opened 2023-07-17
Added the feedback system which replaces the quotecorner and goodideas

ysbrandB:
Complete changed the goodidea board and quotecorner:

- Made the neccesary tables for the feedback systems
- Added feedback categories so every 'feedbackboard' has its own category
- Added the admin tools for the categories
- Added the ability for the board to reply to a feedback, accept or reject it and automatically send an email with a response
- Add the feature that the category can have a reviewer and not publish the feedback until it is approved
- Add a cron job and an email that reminds the reviewer if there are any unreviewed ideas.
- let the old routes of /goodideas and /quotes redirect to /feedback/goodideas and /feedback/quotes

The removal of the quotesystem and goodideaboard is done in a seperate PR
fixes https://github.com/saproto/saproto/issues/1252, fixes https://github.com/saproto/saproto/issues/1181, fixes https://github.com/saproto/saproto/issues/1232

PR #1971, opened 2023-07-17
Remove legacy good idea and quotecorner code and tables

ysbrandB:


ysbrandB:
Made redundant by #1970

PR #1976, opened 2023-07-20
Add back in the correct fields for the event_overview query

ysbrandB:
This code has already been approved in #1821.
I believe in a merge conflict with the laravel 10 branch accidentally this code was reset, making the activity overview useless, as it currently has half the new and half the old code.
This correctly sets it to the new working state again.


Jessseee (approved):
LGTM

PR #1980, opened 2023-08-01
revert the image() relation of companies back to the correct db field

ysbrandB:
With the fixing of updating and creating companies (#1972), my local DB was accidentally seeded on the photo-update branch.
I fixed the updates but accidentally broke the belongsTo relation by setting it to the new Photo field instead of the image field. This breaks the company images and therefore the logo bar. Now reverted.


Menkveld-24 (approved):
Lgtm

Issue #1636, opened 2022-07-07
Combine the news and weekly functionality


ysbrandB:
As the news items on the site are severely underutilised and there is currently no archive of the weekly's I would suggest combining the two. Making the weeklys automatically archive into the news items and removing the distinction between the two and then adding a button below the weekly called 'view older news!'

ysbrandB:
To do: 
- [ ] add is_weekly to newsItem and column
- [ ] Create a many to many table between newsItems and events
- [ ] Rewrite the cronjob
- [ ] Create a new admin view
- [ ] Split the overview in a weekly and non weekly column

Issue #1138, opened 2020-05-04
News letter  archive


Frankieknarf:
I wanted to read an old newsletter from the site, but this wasn't possible because only the latest newsletter is visible. I couldn't find an archive of these on the site. Could this be made possible with a 'see olders news' button?

Thanks!

Jessseee:
With "newsletter" you are referring to the weekly update?

jonathanjuursema:
If you mean the weekly update, unfortunately we don't save these. The functionality that is now used for the weekly update was never meant for the way it is now used, so unfortunately there is no archive for that. :(

Frankieknarf:
yeah, I meant weekly update. I thought it would be nice to add it because it slowly adds historical data to the association. the longer we have these kind of newletters saved, the more we can use them during lustra and stuff like that. It should be low priority however :P

xadya:
Ugly way around this is of course to create an account with a mail connected to it and to save these mails in an inbox

jonathanjuursema:
Also @Frankieknarf see your e-mail. :)

Frankieknarf:
Haha, Thanks for the momentary solution!

PR #1983, opened 2023-08-19
Combine the news and weekly functionality

### LINKED ISSUE #1732: Weekly Newsletter needs to be saved before being able to preview
It's kind of awkward that you currently have to save the weekly newsletter (therefore showing it on the homepage) before you're able to preview it. It would be nicer if when you press the preview button it would use the text inside the text field instead of the already saved text.

ysbrandB:
Combines the weekly functionality into the news functionality.
This way we will have a backlog of weekly overviews.
This also means we can now have a featured image in the weekly mail.
Also added a many to many relationship between news and events. This means that you can add the events to weeklies like u used too, but also to newsarticles (which will show up in a bar at the side). 

Fixes #1732, #1636 and #1138.

Newsadmin: 
image

Newsoverview:
image

Weekly edit:
image

Newsitem edit:
image


Jessseee:
Why not merge it even further? Every weekly update could just be mixed in with the news articles instead of having them still separated as two different things.

ysbrandB:
> Why not merge it even further? Every weekly update could just be mixed in with the news articles instead of having them still separated as two different things.

What would you merge further? The distinction in DB etc is necessary for sending the mails with the weekly and the differing views we use. In my personal opinion the distincion in the overview is also a nice to have.

Jessseee:
I just feel like the overview page would just look weird because the news articles are not written frequently enough. So you will have one empty card for the news articles and a full card for the weekly newsletters, or at least one card with very outdated news articles. Maybe you could mix the two just in the UI. Then there is no need for the models to be merged, but it still makes for a more concise and up-to-date overview.

Jessseee (approved):
LGTM

Jessseee (changes requested):
LNGTM

Jessseee (approved):
LGTM

PR #1985, opened 2023-08-24
Fixed non-members getting orderlines for tickets that never get resolved

### LINKED ISSUE #1984: Potential for giving away free tickets
When a ticket is set to also be available to non-members, but is set to not be prepaid, the system does not check if the account is a member. This means it generates an orderline for a non-member and gives them the ticket, even though we do not have the SEPA of non members and therefore never can resolve the orderline. I believe the behaviour should be checking if the account is a member and  then generate the orderline, otherwise redirect them to the direct payment. Then also rename the option "should be prepaid" to reflect that this option is only for members, like "members should also prepay" or something along those lines.

ysbrandB:
fixes #1984.
Makes it so non-members always have to prepay their ticket, and get sent to mollie.

### DIFF
--- a/resources/views/event/display_includes/tickets.blade.php
+++ b/resources/views/event/display_includes/tickets.blade.php
@@ -91,7 +91,7 @@
 
                                 <p class="card-title">
 
-                                    @if ($ticket->is_prepaid || !Auth::user()->member)
+                                    @if ($ticket->is_prepaid || !Auth::user()->is_member)
                                         @php
                                             $has_prepay_tickets = true;
                                         @endphp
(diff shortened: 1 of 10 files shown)


resources/views/event/display_includes/tickets.blade.php
```diff
@@ -91,7 +91,7 @@ class="card-link text-info">
 
                                 <p class="card-title">
 
-                                    @if ($ticket->is_prepaid)
+                                    @if ($ticket->is_prepaid || !Auth::user()->member)
```
Jessseee, line 94: This should be `user()->is_member` instead of checking for the member model.

Jessseee (changes requested):

PR #1986, opened 2023-08-24
Optimize/omnomcom admin

ysbrandB:
With the orderline wizard we load ALL members and ALL products only for a dropdown (in a wizard you hardly ever see). The wizard is made redundant anyways, as you can now do multiple products over multiple users in the addone quick toolie. I did add the optional description to the addone toolie to not lose functionality.
Also optimized the DB queries.
The page now loads in approximately 350ms versus 4300ms before. (with the same functionality and products shown)
Before: 
image
Now:
image


JonaMata (changes requested):
The orderline wizard is definitely not redundant!
It can be used to add orderlines with custom prices and add a different set of products to multiple users as as well as adding custom descriptions for orderlines.
The treasurer will use this and this should not be removed.

Jessseee:
I think you are trying to solve the wrong problem. The user selection dropdown can definitely be improved in performance, maybe using caching or some other optimization.

ysbrandB:
> I think you are trying to solve the wrong problem. The user selection dropdown can definitely be improved in performance, maybe using caching or some other optimization.

After discussing with Jonathan we came to the conclusion that simply moving the orderline wizard to a new page would solve the issue (as you visit the orderline overview page way more than the wizard). I will open a new PR to make it this way.

PR #1992, opened 2023-09-04
Change "the ticket be visible" to "the ticket WILL be visible"

ysbrandB:
Corrects the english sentence when a ticket will show names on the event page.


Jessseee (approved):
The ticket do be visible. LGTM

PR #1997, opened 2023-09-08
Move the orderline wizard to a new page

ysbrandB:
Moved the orderline wizard to a seperate page.
This means we do not load all the users and products just for the orderline overview.
Also added some querybuilder optimizations to the orderline overview.
Greatly reduces loading time of both (from over 2000ms for the orderline page to both only being 200-300ms)


Jessseee (approved):
LGTM

PR #2006, opened 2023-09-21
Fix the ticket not being deleted when a non member does not pay correctly

ysbrandB:
When the mollie webhook comes back with a failed payment when the ticket was not set to prepaid the ticket would not get deleted for non users. Now fixed.


Jessseee (approved):
LGTM

PR #2011, opened 2023-09-25
Fix the review feedback mails going to the wrong user

ysbrandB:
Also added it so sysadmins can see and edit the unreviewed feedback.


Menkveld-24 (approved):
LGTM

PR #2021, opened 2023-10-03
Feature/event statistics

ysbrandB:
Adds a new query for event/activity statistics.
This automates the overview for the GMM and adds a fun graph with the number of activities over time per board.
The graph it automates is the following:
image

### DIFF
--- a/app/Http/Controllers/QueryController.php
+++ b/app/Http/Controllers/QueryController.php
@@ -2,7 +2,10 @@
 
 namespace App\Http\Controllers;
 
+use App\Models\Activity;
+use App\Models\ActivityParticipation;
 use App\Models\Event;
+use App\Models\EventCategory;
 use App\Models\Member;
 use Carbon\Carbon;
 use DB;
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/QueryController.php
```diff
@@ -157,4 +162,57 @@ public function membershipTotals(Request $request)
...
+                    $query->where('category_id', $category->id)->where('start', '>=', $start)->where('end', '<=', $end);
+                })->where('participants', '>', 0);
+            })->count();
+
+            $category->attendees = Activity::where('participants', '>', 0)->whereHas('event', function ($query) use ($category, $start, $end) {
+                $query->where('category_id', $category->id)->where('start', '>=', $start)->where('end', '<=', $end);
+            })->sum('attendees');
+        }
+
+        $events = DB::table('events')
```
Jessseee, line 203: Why use raw SQL here instead of Eloquent query builders?

ysbrandB: Because the dates are in the wrong format in the DB, and the only way to group by what I want is to use the built-in sql YEAR() and MONTH() functions.

JonaMata: Use the model and then use selectRaw to do your magic sql stuff. Like Event::query()->selectRaw('year(created_at) as year')->groupBy('year')->get()

Wilkuu-2 (approved):
LGTM

PR #2022, opened 2023-10-05
Fix the most upvoted feedback and add the ability to be able to set  publisher visibility

ysbrandB:
Title says it all :)


resources/views/feedbackboards/mostvoted.blade.php
```diff
@@ -2,7 +2,7 @@
 
     <div class="card mb-3">
         <div class="card-header bg-dark text-white">
-            Most upvoted idea
+            Most upvoted recent {{ strtolower(str_singular($category->title)) }}
```
Menkveld-24, line 5:
```suggestion
            Most upvoted recent {{ Str::lower(Str::singular($category->title)) }}
```

I think those methods are the 'old' methods, these are used in the docs of laravel:
https://laravel.com/docs/10.x/strings#method-str-singular

resources/views/feedbackboards/include/feedback.blade.php
```diff
@@ -83,10 +86,11 @@
         <div class="text-muted text-end mt-2">
             <em>
                 <sub>
-                    @can('board')
-                        By {{ $feedback->user?->name ?? "before we kept track!" }}
-                    @endcan
-                    -- {{ $feedback->created_at->format("j M Y, H:i") }}
+                    @if(Auth::user()->can("board") || $feedback->category->show_publisher)
+                        By {{ $feedback->user?->name ?? "before we kept track!" }} --
```
Menkveld-24, line 90:
```suggestion
                       @if(!empty($feedback->user?->name))
                           By {{ $feedback->user->name }} --
                       @else
                           Before we kept track --
                       @endif
```

I think it's a little neater if you don't get 'By before we kept track!' but 'Before we kept track'
But that's just my opinion on it

resources/views/feedbackboards/include/feedback.blade.php
```diff
@@ -44,37 +45,39 @@
...
-                    </form>
-                </div>
-            @endif
+            <div class="collapse mt-3" id="feedback__{{ $feedback->id }}__collapse">
+                <form method="post" action="{{ route('feedback::reply', ['id' => $feedback->id]) }}">
+                    {{ csrf_field() }}
+                    <label for="feedback__{{ $feedback->id }}__reply">Reply:</label>
+                    <textarea id="feedback__{{ $feedback->id }}__reply" class="form-control mb-2" rows="2" cols="30"
+                              name="reply"
+                              placeholder="A reply to this {{ strtolower(str_singular($feedback->category->title)) }}."
```
Menkveld-24, line 65: I think those methods are the 'old' methods, these are used in the docs of laravel:
https://laravel.com/docs/10.x/strings#method-str-singular

resources/views/feedbackboards/include/feedback.blade.php
```diff
@@ -44,37 +45,39 @@
...
+        {!! $feedback->feedback !!}
 
-            @if ($feedback->reply)
-                <hr>
-                <i class="me-1 fa {{$feedback->accepted ? "fa-circle-check text-primary":"fa-circle-xmark text-danger"}}" aria-hidden="true"></i>
-                <b>Board:</b> {!! $feedback->reply !!}
-            @endif
+        @if ($feedback->reply)
+            <hr>
+            <i class="me-1 fa {{$feedback->accepted ? "fa-circle-check text-primary":"fa-circle-xmark text-danger"}}"
```
Menkveld-24, line 52:
```suggestion
            <i class="me-1 fa {{ $feedback->accepted ? "fa-circle-check text-primary" : "fa-circle-xmark text-danger" }}"
```

resources/views/feedbackboards/include/feedback.blade.php
```diff
@@ -44,37 +45,39 @@
     </div>
 
     <div class="card-body">
-            {!! $feedback->feedback !!}
+        {!! $feedback->feedback !!}
```
Menkveld-24, line 48: I know that this was already here, but are we sure that this feedback is safe?

resources/views/feedbackboards/categories.blade.php
```diff
@@ -42,6 +43,12 @@
                                    'label' => 'Should the board be able to reply to this feedback?'
                                ])
 
+                        @include('components.forms.checkbox', [
+                                  'name' => 'show_publisher',
+                                  'checked' => $cur_category->show_publisher??false,
```
Menkveld-24, line 48:
```suggestion
                                  'checked' => $cur_category->show_publisher ?? false,
```

resources/views/feedbackboards/categories.blade.php
```diff
@@ -32,7 +32,8 @@
                         <div id="reviewer" class="{{$cur_category?->review?"":"d-none"}}">
                             <label for="user_id">Reviewer Name:</label>
                             <div class="form-group autocomplete">
-                                <input class="form-control user-search" value="{{$cur_category->reviewer_id??''}}" id="user_id" name="user_id"/>
+                                <input class="form-control user-search" value="{{$cur_category->reviewer_id??''}}"
```
Menkveld-24, line 35:
```suggestion
                                <input class="form-control user-search" value="{{ $cur_category->reviewer_id ?? '' }}"
```

Menkveld-24 (commented):
Just a few remarks, haven't locally tested the code though

resources/views/feedbackboards/mostvoted.blade.php
Jessseee: Yes, we use a Laravel first party library that adds backwards compatibility for these functions so we don't have to replace all these method calls in the code. We should go over the whole codebase some time to fix this.

resources/views/feedbackboards/include/feedback.blade.php
Jessseee: No and also we should not want to render html in the feedback anyways. The feedback should only be plain text in my opinion.

resources/views/feedbackboards/include/feedback.blade.php
ysbrandB: This is also what we do at the membercards so I just wanted to keep it consistent

resources/views/feedbackboards/include/feedback.blade.php
ysbrandB: Yeah I agree, fixed now

Jessseee (approved):
LGTM

PR #2028, opened 2023-10-10
Move Mollie monthly overview to use dates between first mondays of mo…

### LINKED ISSUE #2027: Mollie payment automatisation
Currently in the tab for mollie payments in the Proto site there is an overview of all mollie payments of that month plus the bookkeeping accounts where they need to be booked. But the actualy payouts from mollie are not necciseraly from the first of the month untill the last of the month since payouts are not done on weekends. Is it possible to change the account overview to include weekends if the month ends on a weekend.

JonaMata:
…nths

Fixes #2027


ysbrandB (approved):
LGTM

PR #2034, opened 2023-10-10
Exclude current weekly from news tab + optimize

Wilkuu-2:
The weekly is displayed only in the weekly tab (archived weeklies are still displayed in news)

Also the birthday query is optimized


ysbrandB (approved):
LGTM

PR #2043, opened 2023-10-25
Add rotating carrot to dropdown and small css fixes in the remaining outstanding column

ysbrandB:
This makes it clearer that there is a dropdown and loses the ugly dot in front.
new:
image

### DIFF
--- a/resources/views/omnomcom/orders/includes/payment-details.blade.php
+++ b/resources/views/omnomcom/orders/includes/payment-details.blade.php
@@ -37,40 +37,48 @@
         @endif
 
         @if($outstandingAmount>0)
-            <p class="card-text">
-                Remaining outstanding
-                <i class="fas fa-info-circle" data-bs-toggle="tooltip" data-bs-placement="right" title="The amount caught up in activities you signed up for that still have to be processed. This will not be in the next withdrawal AND you CAN NOT pay this now, but you will have to pay this this money somewhere in the future."></i>
-            </p>
-            <ul class="list-group list-group-flush" id="outstanding-accordion">
+            <div class="border-bottom-0">
+                            <span class="w-100 d-inline-flex justify-content-between">
+                                <span class=" cursor-pointer" data-bs-toggle="collapse"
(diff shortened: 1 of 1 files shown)


resources/views/omnomcom/orders/includes/payment-details.blade.php
```diff
@@ -37,40 +37,48 @@
...
-                    <div id="outstanding" class="collapse" data-parent="#outstanding-accordion">
-                        <table class="table table-borderless table-hover table-sm mt-1">
+            <div class="border-bottom-0">
+                            <span class="w-100 d-inline-flex justify-content-between">
+                                <span class=" cursor-pointer" data-bs-toggle="collapse"
+                                      data-bs-target="#collapse-outstanding">
+                                     Remaining outstanding
+                                    <i class="fas fa-info-circle" data-bs-toggle="tooltip"
+                                       data-bs-placement="right"
+                                       title="The amount caught up in activities you signed up for that still have to be processed. This will not be in the next withdrawal AND you CAN NOT pay this now, but you will have to pay this this money somewhere in the future."></i>
```
Jessseee, line 47: This sentence is not great. Notice the double "this" in the horribly long second sentence. Maybe replace with something more concise or at least more well structured. However, I generally think this "remaining outstanding" amount is terribly confusing.

Jessseee (changes requested):

resources/views/omnomcom/orders/includes/payment-details.blade.php
ysbrandB: What would you suggest to make this more clear?

ysbrandB: Would this be better: "he amount owed for activities you signed up for that have not yet been processed by the treasurer. Therefore, you can not already pay this."?

ysbrandB: We have solved this in the meeting.

PR #2045, opened 2023-10-26
Bringing back the december christmas theme!

ysbrandB:
As requested by the board. Only available in December and togglable via cookie.

### DIFF
--- a/resources/assets/js/application.js
+++ b/resources/assets/js/application.js
@@ -1,10 +1,13 @@
 window.global ||= window;
 // Vendors
 import SignaturePad from "signature_pad";
+
 global.SignaturePad = SignaturePad;
 import moment from "moment/moment.js";
+
 global.moment = moment;
 import quagga from 'quagga';
+
 global.Quagga = quagga;
 
 import './countdown-timer'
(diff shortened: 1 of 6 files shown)


resources/assets/js/application.js
```diff
@@ -1,34 +1,44 @@
...
+
 global.Quagga = quagga;
 
 import './countdown-timer'
 import './utilities'
 import './broto'
 import './nightMode'
 // Execute theme JavaScript
-window[config.theme]?.()
+if ((new Date().getMonth() + 1) !== 12) {
```
Jessseee, line 18: Why add this check here?

ysbrandB: Otherwise you will still get all the stars from nightmode and all the brotos from the broto theme, it would be best if this check could know the theme but I do not think we have acces to that in the JS only files.

ysbrandB: No the needed acces is to the disable-december cookie but that does not get to the js. When I console.log(document.cookie) it does not show up, even when set

Jessseee: Sure, I guess we revert this code after December anyways.

PR #2047, opened 2023-11-02
Fix the linebreaks showing up in feedback

ysbrandB:
should also run the sql query: `UPDATE feedback SET feedback=REPLACE(feedback, '<br />', '') WHERE 1`


JonaMata (approved):
LGTM

PR #2050, opened 2023-11-07
Replace mt_rand by more random random_int method

### LINKED ISSUE #2049: Protube not random enough
On protube, some pictures from certain albums get the preference of the algorithms. I don't know how or why but it's just not really random. I suspect it has a preference of more "recent" photos in albums....

ysbrandB:
Might Fix #2049.


Jessseee (approved):
LGTM

PR #2051, opened 2023-11-07
Allow admins to remove user accounts

Jessseee:
See title.

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -237,6 +237,7 @@ class AuthController extends Controller
 
         if ($user->name != $request->name) {
             Session::flash('flash_message', 'You need to correctly input the user\'s name before the account is deactivated.');
+
             return Redirect::back();
         }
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/AuthController.php
```diff
@@ -233,25 +233,36 @@ private function registerAccount($request)
...
+
+        if ($user->hasUnpaidOrderlines()) {
+            Session::flash('flash_message', 'An account cannot be deactivated while it has open payments!');
+
+            return Redirect::route('omnomcom::orders::list');
         }
 
         if ($user->member) {
-            Session::flash('flash_message', 'You cannot deactivate your account while you are a member.');
+            Session::flash('flash_message', 'An account cannot be deactivated while it still has an active membership.');
```
ysbrandB, line 265: I think we should also redirect them to somewhere usefull here (or a message how to deactivate their membership) like we do at the unpaidOrderlines

ysbrandB (changes requested):

ysbrandB (approved):
LGTM

PR #2052, opened 2023-11-08
Fix stars in achievement gallery

EmmaBurema:


resources/views/achievement/gallery.blade.php
```diff
@@ -8,7 +8,7 @@
 
     <div id="achievement-accordion">
 
-        <?php $stars = 1; ?>
+        <?php $stars = 0; ?>
```
JonaMata, line 11: This variable seems totally unnecessary, I think it can be removed (including the ++ at the end)

JonaMata (changes requested):
LAGTM

JonaMata (approved):
LGTM

PR #2053, opened 2023-11-14
E-mail theme fix

lucer1:
Now the weekly's are in the user's selected theme. The e-mails can't select the user's theme (something with cookies i heard), but i checked and all the e-mails still function properly.

Next to that, I fixed some issues in the theme files, because on the 'your profile' page, some bits couldn't be read because of wrong colors.

And lastly, I saw some spelling mistakes and fixed those.


ysbrandB (approved):
LGTM

PR #2055, opened 2023-11-23
Feature/Added inactive committees

Frank-Bosman:
# Added Inactive committees and societies
These inactive committees/societies are shown separately on the overview page.

## Edit/new committee page
afbeelding

## Societies overview page
afbeelding
For the committees page it's the same, but the societies is changed to committees.


ysbrandB (approved):
LGTM

PR #2057, opened 2023-11-30
Feature/protube cleanup v2

Menkveld-24:
https://github.com/saproto/ProTube/pull/134 of Protube is recommended for this update.

Removes all old protube stuff in the site. The soundboard is kept as we might use that part in the future (so we keep the current files).
Implemented the skipsong endpoint and tempadmin endpoint.

Note: .env needs updating on production.
HERBERT_SERVER -> PROTUBE_SERVER
HERBERT_SECRET -> PROTUBE_SHARED_SECRET


ysbrandB (approved):
LGTM

PR #2064, opened 2023-12-18
Restore tempadmin functionality

Menkveld-24:
- Restored the accidentally deleted button
- Updated the isTempadminLaterToday function to check if the user is tempadmin between now - end of today.
- Added update calls for the protube api.


ysbrandB (approved):
LGTM

PR #2066, opened 2023-12-19
Hide exception info on 500 error page

JonaMata:


ysbrandB (approved):
Lookily goodily to me

PR #2082, opened 2024-01-25
Show the error message when the environment is local

ysbrandB:
When developing on local the message when impersonating a normal user also was hidden. Will now be shown

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -60,7 +60,7 @@ class Handler extends ExceptionHandler
     /**
      * Render an exception into an HTTP response.
      *
-     * @param  Request  $request
+     * @param Request $request
      * @return SymfonyResponse
      *
      * @throws Throwable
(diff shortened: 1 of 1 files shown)


app/Exceptions/Handler.php
```diff
@@ -97,7 +97,9 @@ protected function unauthenticated($request, AuthenticationException $exception)
     protected function renderHttpException(HttpExceptionInterface $e)
     {
         if (! view()->exists("errors.{$e->getStatusCode()}")) {
-            return response()->view('errors.default', ['exception' => $e, 'hide_message' => (! Auth::check() || Auth::user()?->canNot('finadmin'))], 500, $e->getHeaders());
+            $maySeeError = App()->env('local') | Auth::check() && Auth::user()->can('finadmin');
```
Jessseee, line 100: maybe use `||` and shouldn't `finadmin` just be `admin` or `sysadmin`

Jessseee (changes requested):

app/Exceptions/Handler.php
ysbrandB: No, due to a bug in our financial system the treasurer should be able to also see the errors until it is fixed. See last minutes

Jessseee (approved):
LGTM

PR #2085, opened 2024-01-25
Remove all helper reminder emails

ysbrandB:
As discussed in the meeting of today, these are pretty much useless as all this communication gets done via directer means of communication. We have had more problems with these then they ever helped, so it is time to let go.


Wilkuu-2 (approved):
LGTM, 
Pint does not approve of your formatter, though.

PR #2088, opened 2024-01-26
Fixed the IBAN check and update the recognized IBAN list based upon the SEPA list

ysbrandB:
Fix the IBAN check being able to pass when a dutch IBAN is used but the identifier is not in our table.
Also updates the list of recognized BIC codes based upon the list of the betaalvereniging
https://www.betaalvereniging.nl/wp-content/uploads/BIC-lijst-NL.xlsx

I did check in the database if there were BICS used which are now out of the list and there were none so we can delete these safely, as they are no longer SEPA supported. 
Those being:
'AABNNL2A',
'ATBANL2A',
'BCDMNL22',
'BMEUNL21',
'DLBKNL2A',
'KASANL2A',
'LOCYNL2A',
'MOYONL21'


Jessseee (dismissed):
LAGTM

Jessseee (approved):
LGTM

PR #2089, opened 2024-01-26
Update the SetTimeOut to SetInterval and make the times longer

ysbrandB:
This is done for the protopolisscreen, narrowcasting and smartxp screen


Jessseee (approved):
LGTM

PR #2092, opened 2024-02-03
Fix badMethodCall to app()->env. It is App::environment

ysbrandB:
Whoops, my lsp showed this as a function and because I have the rights still got to see the error.


Jessseee (approved):
:(

PR #2101, opened 2024-02-16
Add the ability for the board to change another users email

ysbrandB:
Adds a new modal in the user admin for the board to change someone elses mail.
They need to input their (the boards) password as confirmation.
I have tested it as board for myself and for someone else. Also as not a board.
This needs to be tested thoroughly by others though for safety.
The new modal:

image


resources/views/users/admin/admin_includes/contact.blade.php
```diff
@@ -7,7 +7,13 @@
     <div class="card-body">
 
         <p class="card-text">
-            <i class="fas fa-at fa-fw me-2"></i> <a href="mailto:{{ $user->email }}">{{ $user->email }}</a><br>
+            <i class="fas fa-at fa-fw me-2"></i> <a href="mailto:{{ $user->email }}">{{ $user->email }}</a>
+            @if($user->id !== Auth::user()->id)
```
Jessseee, line 11: Maybe also add a "can board" check, just to be sure .

Jessseee (changes requested):
LAGTM

resources/views/users/admin/admin_includes/contact.blade.php
ysbrandB: By reading this again I now also see a board member could update a sys admins email and get that account for themselves. I will add a check like we have with the impersonation to make it so you can only change the email from someone with the same or lower amount of rights as you.

Wilkuu-2 (approved):
Tested it, a potential board-exclusive exploit is patched out. 
LGTM

PR #2102, opened 2024-02-18
Fix the display helpers being shown when not member

ysbrandB:
fixes https://sa-proto.sentry.io/issues/4328386371/?project=1278929&referrer=slack


Jessseee (approved):
LGTM

PR #2104, opened 2024-02-21
Add the ability to dictate how many of a ticket a user is able to buy

### LINKED ISSUE #2012: Add the option for tickets to only be able to buy one per user
This way for for instance camp we will not need to check duplicates

ysbrandB:
For instance for camp you could say a user can buy only one ticket, or maybe you do not want more than two food tickets per user oid. Fixes #2012.


Wilkuu-2 (approved):
LGTM

PR #2108, opened 2024-02-27
Always show participation limits

### LINKED ISSUE #1660: Participant availability for non-open events
Dear HYT,

I think the title says it all, however, it was something that I encountered again this week.
Whenever an event is still closed, you cannot see how many spots are available.
This can be quite annoying for events that may be popular, as you want to know how speedy you want to get your tickets.
May be a nice quick and easy fix.

Wilkuu-2:
Now the limits on participations are visible when the event is not open!

Fixes #1660


Wilkuu-2:
### The limit can be seen here:

image

MaxLiebe (approved):
LGTM, shouldn't conflict with #2119 I think. Otherwise a quick fix.

PR #2110, opened 2024-02-28
Run rector with the early return set

ysbrandB:
This ensures early returns are used when they can be.


Wilkuu-2:
Replaces with something that sucks less (targeted at rector)

PR #2112, opened 2024-02-28
Added question mark to $event->img = $event->image?->generatePath();

ysbrandB:
This makes that if there is no image the query does not break


resources/views/wallstreet/admin_includes/wallstreetdrink-event-list.blade.php
```diff
@@ -14,7 +14,6 @@
 
                     <th>ID</th>
                     <th>Title</th>
-                    <th>Description</th>
```
Wilkuu-2, line 17: Isn't emoving the description from the overview a separate change?

Wilkuu-2 (approved):
LGTM, You should not sneak in additional changes without changing the title...

PR #2114, opened 2024-03-04
Add the wallstreet event image_id to the Storage entry check

ysbrandB:
This is to prevent it from being deleted by our cleanup cron.


Frank-Bosman (approved):
LGTM

PR #2118, opened 2024-03-12
Fix: Remove unnecessary check that counts events per year

ysbrandB:
This was only being done to ensure the years have events, but the original years call only gets years where there are events anyway. Saves about 10ms/year we have events in.

The query where we get the years already ensures that only years are gotten which have events in them:
image


Wilkuu-2 (approved):
LGTM

PR #2119, opened 2024-03-12
Fix/event page performance

ysbrandB:
Fixed a lot of performance issues on the event page. This makes getting an archived year or the event overview page about 11x faster. 
image

Went from around 8,5 seconds just in the DB with 2000 models loaded:
image

To around 270ms in the DB with 570 models loaded. 
image


resources/views/event/display_includes/event_block.blade.php
```diff
@@ -107,20 +102,20 @@
             @endif
 
             {{-- Signup Icon --}}
-                <div class= "d-flex justify-content-between">
-                    @if($event->usersCount()>0)
-                        <span>
+            <div class="d-flex justify-content-between">
+                @if($event->unique_users_count>0)
```
Menkveld-24, line 106: Almost

resources/views/event/display_includes/event_block.blade.php
```diff
@@ -107,20 +102,20 @@
...
+                    <span>
                             <i class="fas fa-user-alt fa-fw" aria-hidden="true"></i>
-                            {{$event->usersCount()}}
+                            {{$event->unique_users_count}}
```
Menkveld-24, line 109: .

resources/views/website/home/members.blade.php
```diff
@@ -48,7 +48,7 @@
 
     <div class="col-xl-4 col-md-12">
 
-        @include('website.home.cards.upcomingevents', ['n' => 6])
+        @include('website.home.cards.upcomingevents', ['events'=>$upcomingEvents])
```
Menkveld-24, line 51: .

Menkveld-24 (changes requested):

resources/views/committee/include/organised-events.blade.php
```diff
@@ -56,8 +56,10 @@
     </div>
 
 @endif
-
-@if(count($committee->pastHelpedEvents()) > 0)
+@php
```
MaxLiebe, line 59: I know this limit is already in place in master, but it's kind of weird there is no pagination on the committee page to see more than 6 previously organised events. Especially for a committee like pLAN that organises a lot of actvities, mainly GameBreaks, it can give an unrealistic representation of the committee.

resources/views/event/display_includes/event_block.blade.php
```diff
@@ -24,32 +24,26 @@
             @endif
 
             {{-- Participating --}}
-            @if(Auth::check() && $event->activity?->isParticipating(Auth::user()))
-                @if($event->activity->isOnBackupList(Auth::user()))
+            @if(Auth::check() && $event->activity?->myParticipationCount > 0)
```
MaxLiebe, line 27: Why were these booleans changed to integers exactly? I think the old syntax makes more sense and is more readable. How is someone supposed to have multiple participations and multiple backup list participations?

resources/views/event/display_includes/event_block.blade.php
```diff
@@ -24,32 +24,26 @@
...
+                @if($event->activity->myBackupParticipationCount > 0)
                     <i class="fas fa-check text-warning fa-fw" aria-hidden="true"
                        data-bs-toggle="tooltip" data-bs-placement="top" title="You are on the backuplist!"></i>
                 @else
                     <i class="fas fa-check text-primary fa-fw" aria-hidden="true"
                        data-bs-toggle="tooltip" data-bs-placement="top" title="You are participating!"></i>
                 @endif
-                @if($event->activity->isHelping(Auth::user()))
+                @if($event->activity->myHelperParticipationCount > 0)
```
MaxLiebe, line 35: Same here tbh, old syntax is definitely clearer

resources/views/event/display_includes/event_block.blade.php
```diff
@@ -24,32 +24,26 @@
...
             @endif
 
             {{-- Ticket --}}
-            @if (Auth::check() && $event->hasBoughtTickets(Auth::user()))
+            @if (Auth::check() && $event->myTicketCount>0)
                 <i class="fas fa-ticket-alt fa-fw text-info" aria-hidden="true"
                    data-bs-toggle="tooltip" data-bs-placement="top" title="You bought a ticket!"></i>
             @endif
 
-            {{-- Helper --}}
```
MaxLiebe, line 47: ??? why was this removed

MaxLiebe (changes requested):
Willing to accept the review if you can give me a good explanation why the booleans were changed to integers and why the "This activity needs your help!" reminder was removed.

resources/views/committee/include/organised-events.blade.php
ysbrandB: Would be a nice feature request in a new issue :)

resources/views/event/display_includes/event_block.blade.php
ysbrandB: This should have been removed when we removed the helper reminder emails because we do not use this system anymore

resources/views/event/display_includes/event_block.blade.php
ysbrandB: The old syntax did extra database calls for every rendered activity. Now it is loaded upfront. I could move this check to the controller, but then this syntax would just be moved. We need to know per activity if you have any tickets or participation, and counting them in the same query is the most performant vs running 4 or 5 queries per activity.

Menkveld-24 (approved):
Gut gut

PR #2122, opened 2024-03-22
Optimization: lazy load background images

ysbrandB:
Lazy load images as described in: https://www.conroyp.com/articles/improve-load-time-performance-lazily-loading-background-images

This will prevent us from loading all images at once, for instance in the calendar overview. For now, only implemented for the event_block. This means those images only get loaded when the IntersectionObserver notices the image is within 100 pixels of coming into view.


Jessseee (approved):
LGTM

PR #2124, opened 2024-04-05
WIP: Add website smtp account to DA sync

JonaMata:


app/Console/Commands/DirectAdminSync.php
```diff
@@ -308,6 +317,21 @@ private function applyAccountPatchList($patch)
...
+        foreach ($patch['add_system'] as $account) {
+            $queries[] = [
+                'cmd' => '/CMD_API_POP',
+                'options' => [
+                    'domain' => getenv('DA_DOMAIN'),
+                    'action' => 'create',
+                    'user' => $account,
+                    'passwd' => config('mail.password'),
+                    'passwd2' => config('mail.password'),
+                    'quota' => 0, // Unlimited
```
ysbrandB, line 329: These shouldn't be unlimited but a sensible default.

ysbrandB (changes requested):

app/Console/Commands/DirectAdminSync.php
JonaMata: I think for the system account setting it to unlimited is fine. Maybe for the normal users we can set lower limits.

PR #2125, opened 2024-04-08
Attempt to fix the spotifysync

ysbrandB:
The addPlaylistTracks function does not take the Spotify user as an argument. Therefore it tried to use the Spotify user as the base 62 playlist id which breaks. By giving it the arguments it expects (playlist id and songs) it should work again.


Wilkuu-2 (approved):
LGTM

PR #2126, opened 2024-04-08
Fix the protube music history

ysbrandB:
Add the new env variable for the protube sync that protube expects for calling the laravel api endpoint. It is already in the live .env file so can be merged directly.


config/protube.php
```diff
@@ -8,6 +8,6 @@
 
     'server' => env('PROTUBE_SERVER'),
     'secret' => env('PROTUBE_SHARED_SECRET'),
```
Jessseee, line 10: Could you rename this to `shared_secret`?

Jessseee (changes requested):

Jessseee (approved):
LGTM

PR #2133, opened 2024-04-16
Install laravel reverb (php based websockets) for dev and use ably for prod and move the wallstreet event updates to it

ysbrandB:
Followed the installation guide at https://laravel.com/docs/11.x/reverb.
To test moved over the wallstreet drink updates to a private websocket channel.
Instead of polling the page every x seconds it will just receive an update once the events are fired, and update the prices and the charts accordingly

For production we use the ably free tier. https://ably.com/


.env.dev
```diff
@@ -94,4 +92,27 @@ HASHIDS_SECRET_EVENTS=RAND_STRING
...
-SEPA_CI=CREDITOR_ID
\ No newline at end of file
+SEPA_CI=CREDITOR_ID
+
+BROADCAST_DRIVER=reverb
+BROADCAST_CONNECTION=reverb
+
+REVERB_APP_ID=959336
+REVERB_APP_KEY=cu7qbxxhf3hrfqpfru1h
+REVERB_APP_SECRET=urrws04xxsb1rehkf8nc
```
MaxLiebe, line 102: YSBRAND!!!

(this shouldn't be public right??!!)

README.md
```diff
@@ -123,6 +123,18 @@ for changes to scripts or stylesheets.
 When adding a new library or client-side dependency through npm don't forget to require the scripts in `application.js`
 and the stylesheet in `vendor.scss`.
 
+#### Websockets
+
+In some parts of the website we use websockets to update the page in real-time.
+In development, we use the [Reverb](https://reverb.laravel.com/) websocket server.
```
MaxLiebe, line 129: Why do we not use Reverb for production as well? Using two different websocket servers seems convoluted.

MaxLiebe (changes requested):
Cannot accept this yet. This definitely needs to be discussed at a meeting first.

README.md
ysbrandB: Reverb does not play nice with our webhosting. Since the two communicate via the laravel adapter there are no differences between using the two. For local it is way easier using reverb, and for prod we should not be going over the ably free tier.

PR #2134, opened 2024-04-18
Add that everyone in the proto.rootcommittee (HYT) should have 2fa enabled.

ysbrandB:
They have more acces to data via the exports endpoint etc. Therefore we think they should also enable 2FA.


Jessseee (approved):
LGTM

Jessseee (approved):
Whoops, LGTM

PR #2136, opened 2024-04-19
Add the weizen big to the true german achievement

ysbrandB:


Frank-Bosman (approved):
LGTM,
Tested it and it works. Do need 21 weizen instead of 20, but that's another issue #2137.

PR #2139, opened 2024-05-05
Fix sail alias in installation instructions for macOS users

MaxLiebe:
Apple switched from using bash for shells to using zsh in macOS Catalina, so the old command doesn't work anymore. Added a new command that does the same but actually works.


ysbrandB (approved):
LGTM

PR #2140, opened 2024-05-05
Set admin account password to be fixed instead of randomly generated

MaxLiebe:
I don't know why this password is randomly generated every time. Seems unnecessary to me. If you want to suggest a better default password than 'proto' be my guest.


ysbrandB (changes requested):
Set a better default password

ysbrandB (approved):
LGTM

PR #2145, opened 2024-05-10
Add the ability to link a Discord account to your Proto account

MaxLiebe:
Used by [ProBoto](https://github.com/saproto/ProBoto) to verify Proto membership status. This functionality has also been implemented into ProBoto now.


MaxLiebe:
Something seems to have duplicated this branch, renaming it to have a capital letter F in its name. I have no idea how that happened. The changes are all correct though so this pull request can be reviewed without any issues.

app/Http/Controllers/ApiController.php
```diff
@@ -253,4 +253,18 @@ public function gdprExport()
 
         return $data;
     }
+
+    public function discordVerifyMember($userId): JsonResponse
```
ysbrandB, line 257: Since this is an publically available endpoint I think it should have some kind of secret/bearer token so you can not just try all discord combinations against our DB

resources/views/users/dashboard/includes/discord_link.blade.php
```diff
@@ -0,0 +1,32 @@
...
+        <p class="card-text">
+            <i class="fa-solid fa-check"></i>  You have linked a Discord account!
+        </p>
+        @endif
+
+    </div>
+
+    <div class="card-footer">
+        @if(!$user->discord_id)
+        <a href="https://discord.com/api/oauth2/authorize?client_id=1088465589099561043&redirect_uri=http%3A%2F%2Flocalhost%3A8080%2Fapi%2Fdiscord%2Flinked&response_type=code&scope=identify" class="btn btn-outline-info btn-block">
```
ysbrandB, line 22: Is this link linked to our discord specifically?

ysbrandB (changes requested):
Not tested yet, but at least should include a secret in the endpoint

resources/views/users/dashboard/includes/discord_link.blade.php
MaxLiebe: Whoops, that should definitely be .env'd. Not sensitive information though so it's not a problem but I'll make sure to make it a config parameter. Also the redirect URL is hardcoded and doesn't use APP_URL which is also stupido. Will fix!

app/Http/Controllers/ApiController.php
MaxLiebe: Yup, good catch

ysbrandB (dismissed):
LGTM

ysbrandB (approved):
LGTM

PR #2148, opened 2024-05-13
Fix the ability to save tickets

### LINKED ISSUE #2147: Tickets cant be changed anymore after they are created :(
When I want to change the end date of a ticket, the site doesn't let me click submit (nothing happens when I do)

ysbrandB:
Setting this to a negative number is dumb but does not break anything. Saves writing unnecessary JS
Fixes #2147


FrankBosman (approved):
LGTM

MaxLiebe:
I'd say LGTM as well but why does this even break lol? (I rather would have some sort of limit/sanitisation in place than nothing at all)

PR #2149, opened 2024-05-13
Fix being able to process an already processed dinnerform

ysbrandB:
Title says it all


MaxLiebe (approved):
LGTM, guessing this prevents fat-fingering the button/double clicking from processing it twice?

ysbrandB:
> LGTM, guessing this prevents fat-fingering the button/double clicking from processing it twice?

Yes, it prevents f5-ing the request accidentally and the processing being done twice

PR #2150, opened 2024-05-14
Feature: reload to last opened tab in the tipcie store

ysbrandB:
Sets the id of the last opened product id in session storage, and on reload if the slug is tipcie, select that tab.
This is usefull for outside drinks, then the drafter does not need to click on the outside tab every time


MaxLiebe:
Can you tell me the line numbers of where a change happened? Because the formatting of the whole file was altered I have no idea what you changed.

ysbrandB:
> Can you tell me the line numbers of where a change happened? Because the formatting of the whole file was altered I have no idea what you changed.

The lines added/changed are 112-130:
``` const categoryBtnList = Array.from(document.getElementsByClassName('btn-category'))
        categoryBtnList.forEach(el => {
            el.addEventListener('click', _ => {
                setTabActive(el)
            })
        })
        let lastSelectedTab = document.querySelector(`[data-id="${localStorage.getItem("currentProductPageId")}"]`);
        if (lastSelectedTab && "{{$store_slug}}" === "tipcie") {
            setTabActive(lastSelectedTab);
        } else {
            setTabActive(categoryBtnList[0]);
        }
        function setTabActive(el) {
            Array.from(document.querySelectorAll('#category-nav > .active')).forEach(el => el.classList.remove('active'))
            el.classList.add('active')
            const categoryViewList = Array.from(document.getElementsByClassName('category-view'))
            const id = el.getAttribute('data-id')
            localStorage.setItem("currentProductPageId", id);
            categoryViewList.forEach(el => {
                if (el.getAttribute('data-id') !== id) el.classList.add('inactive')
                else el.classList.remove('inactive')
            })
        }``` I set the last active tab in local storage if it is the tipcie store and set it active when reloaded

PR #2151, opened 2024-05-15
Fix the search being null when less than 3 characters

ysbrandB:
Fixes https://sa-proto.sentry.io/issues/5320218966/.
The query could be an empty array when there are less than 3 characters, now it returns empty and handles this case


MaxLiebe (approved):
LGTM, now please review my pull request as well

PR #2155, opened 2024-05-22
Fix the migration for the unique users count

ysbrandB:
somehow the column already exists on the live server so added a check that removes it first when exists and then updates it. Adding the check is safer than doing sql queries on live.


Wilkuu-2 (approved):
LGTM

PR #2156, opened 2024-05-22
Chunk the events to avoid our memory limit

ysbrandB:
Loading all events is too much, now chunked in batches of 25


Wilkuu-2 (approved):
Sure, LGTM

PR #2158, opened 2024-05-28
Codex improvements

ysbrandB:
Because songs could have multiple categories attached they could appear in a codex twice (one time under each category). This was barely used and unwanted. Now it has a simple dropdown so it can only have one category. The songs with two categories have already been fixed by hand on the live website. Also the shuffle functionality is removed since that did not do anything.

Added a new command which takes care of all the ?? still left in the codices. Now only select songs still have a ? somewhere. But all the other encoding should be fixed.


Menkveld-24 (changes requested):
afbeelding
This seem so occur on most pages

Menkveld-24:
afbeelding
And maybe we might want to make names required for texts

ysbrandB:
Another rewrite will be done and then the PR reopened.

PR #2159, opened 2024-05-28
Improvement/lazyloading and paginate the dinnerforms

ysbrandB:
Fix the lazy loading of relations on a whole bunch of pages. 
This prevents the blade n+1 problem.
It will be improved in the future for other pages. 
Then I recommend also putting  Model::preventLazyLoading(!app()->isProduction()); in the boot method of our appservice provider to prevent these problems in the future.


Wilkuu-2 (approved):
LGTM ¯\_(ツ)_/¯

PR #2160, opened 2024-05-29
Improvement: Cache our github actions

ysbrandB:
Caches rector, pint and the composer install via their respective actions and the github caching action. The cache folder for this is /.temp.
The composer install options seem changed but the --no-interaction is always passed in the action. [Read here](https://github.com/ramsey/composer-install?tab=readme-ov-file#composer-options)


MaxLiebe (approved):
LGTM

PR #2163, opened 2024-06-03
Fix events without publication time not showing up on the homepage

ysbrandB:
If they did not have a publication time they would not be fetched at all. Now changed to also include the null state.


Jessseee (approved):
LGTM

PR #2170, opened 2024-06-18
Add the Google Workspace Sync

Jessseee:
Implemented a command to synchronise users and committees to the Google Workspace.

This will create Google groups for committees and Google users for active members and add the users to the appropriate groups. It will also add aliases that point at specific users or committees. When a user is created the user's configured email will automatically be added as a forwarder in their Google Workspace Gmail. When a Google user is created the member will first have to use the password sync in their dashboard (it would be nice if users are automatically notified of this).

Also requires Google `application_credentials.json`.


Jessseee:
To test this you should add `->slice(0, 1)` to the `ProtoUser::query()` in `syncProtoUsersWithGoogleUsers()` to not sync all of the seeded users to Google Workspace. Also disable the `!app()->environment('production')` check in `handle()`, or change the environment to production.

.env.prod
```diff
@@ -66,10 +66,6 @@ LDAP_PROXY_KEY=VERY_SECRET_KEY
 
 MOLLIE_KEY=test_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 
-DISCORD_CLIENT=oauth_client_id
```
ysbrandB, line 69: These should probably stay in the .env

database/seeders/DatabaseSeeder.php
```diff
@@ -29,7 +29,7 @@ public function run()
 
         $output->task('setting roles and permissions', fn () => Artisan::call('proto:syncroles'));
 
-        $adminPassword = 'proto';
+        $adminPassword = str_random();
```
ysbrandB, line 32: Should this be changed back?

database/seeders/ImportLiveDataSeeder.php
```diff
@@ -63,7 +63,7 @@ public function run($password, $output)
     public static function getDataFromExportApi($tableName)
     {
         $local_url = route('api::user::dev_export', ['personal_key' => config('app-proto.personal-proto-key'), 'table' => $tableName]);
-        $remote_url = str_replace(config('app-proto.app-url'), 'https://www.proto.utwente.nl', $local_url);
+        $remote_url = str_replace(config('app-proto.app-url'), 'https://www.proto.utwente.nl/', $local_url);
```
ysbrandB, line 66: Also here, should the / be readded here? we do not have it there in production.

ysbrandB (changes requested):

PR #2171, opened 2024-06-18
Fix the event copy copying too much data and add a command to recalculate the unique users count

### LINKED ISSUE #2165: Copied events carry participation(count)s with them.
You cannot sign up for the drink, but there are still people signed up. It seems that the counts are taken from the old event. 

image

ysbrandB:
Fixes #2165 and probably makes the event overview usefull again


Wilkuu-2 (approved):
LGTM

PR #2174, opened 2024-07-05
Hide participants count in the calendar depending on sign up open date

Wilkuu-2:
This commit hides the participant count in the calendar if the sign up date is in the future.


ysbrandB:
Why would we want to hide this? The only scenario where this happens that I can think of is when the board already signs themselves up but I do not think then showing that number is a big deal.

ysbrandB:
I do not think this is normally necessary, and therefore will close it for now. This happened due to a bug with the copying of events. Should not be relevant anymore. Please reopen if you still think it is.

PR #2175, opened 2024-07-08
Re add the discord methods in their own controller

ysbrandB:
These where accidentally deleted with this [commit](https://github.com/saproto/saproto/commit/962e5a5aa59f2ce88463975aaae22c3dbf8596e8). I do not think these methods should have been in the userapicontroller. Now moved to their own DiscordController.


Jessseee (approved):
LGTM

PR #2177, opened 2024-07-10
Improvements/google api

ysbrandB:
Clarify the Gsync, rework the groups (committees) in the models to make that easier. Also add an email that gets sent when a new Workspace account is added.


Jessseee:
Manually merged into other Google API branch.

PR #2180, opened 2024-08-16
Made FontAwesome kit URL configurable in .env

Jessseee:
Could make it easier to quickly switch FontAwesome kit URL if necessary


resources/views/website/assets/javascripts.blade.php
```diff
@@ -19,4 +19,4 @@
 
 @vite('resources/assets/js/application.js')
```
ysbrandB, line 21: Does this not need quotes around it?

Jessseee: You are right.

ysbrandB (approved):
LGTM

PR #2181, opened 2024-08-26
Remove 'Register with Unversity Account' button

Wilkuu-2:
This flow might be broken, so for now it is better to disable it.


ysbrandB:
This does not work with the html comments, made a new PR

PR #2182, opened 2024-08-26
[Archived] Replace fontawesome kit with a npm package in case the fontawesome kit is over its usage

Wilkuu-2:
This should permanently solve our fontawesome woes.

Might increase bandwidth on our end.


Wilkuu-2:
Drafted the PR in case the Font Awesome kit faces problems again.

PR #2183, opened 2024-08-26
Comment out the register with university account button

ysbrandB:
This does not always work and linking does, so for now commented it out


Wilkuu-2 (approved):
LGTM

PR #2184, opened 2024-08-26
Remove all the times env() is called outside of the config files

ysbrandB:
With caching these will return null.
This has not been happening because the double routes prevented caching but will become a problem otherwise.


config/directadmin.php
MaxLiebe, line 1: Might as well change these to env() instead of getenv() for consistency (and to use Laravel's syntaxes)

MaxLiebe (changes requested):
LGTM, just do the smol change I requested and it can be merged

config/directadmin.php
MaxLiebe: Also makes more sense as env() returns null if the variable doesn't exist whereas getenv() returns false

ysbrandB: resolved

PR #2185, opened 2024-08-27
Fixed dark theme navbar link hover bg-color

Jessseee:
Fixed navbar links being unreadable when hovered in dark theme.

**Before:**

image

**After:**

image


ysbrandB (approved):
LGTM

PR #2187, opened 2024-08-29
As discussed in GMM AP 50.15 (Remove the option to become a member through the Discord.)

ysbrandB:
This was the only notion of becoming a member in DC I could find, not even something about it in the discord itself


Jessseee (approved):
LGTM

PR #2188, opened 2024-08-29
Remove the postcode api from the signup process

ysbrandB:
There is no documentation on the error response, breaking our sign-up process when the address was not found. Removing the API removes complexity, and does not take away much. It now automatically prefills the fields with Enschede, The Netherlands which can be edited. This also removes the sentry error 'the street can not be null' as that only happened when the address could not be found.

The new fields look like this: 
image


MaxLiebe (approved):
LGTM

PR #2190, opened 2024-08-31
Replace invite.gg/proto to saproto.nl/go/discord

ysbrandB:
This makes it so we can replace the discord invite link via the url shortener when it changes.


Jessseee (approved):
LGTM

PR #2194, opened 2024-09-04
Temp fix for international bank accounts not being able to register for payment authorisation

### LINKED ISSUE #2193: It is actually theoretically impossible issue a payment authorisation for someone with a non-Dutch bank account
Was reading through the source code that validates IBANs & BICs and discovered something quite amusing: you cannot (without breaking the validation or using inspect element) enter the bank account details of a non-Dutch bank account. 

To understand why this is, note that by default the BIC input field is empty and disabled. To get the option to fill in a BIC, you must first enter an IBAN, which is then validated using BankController's verifyIban() method. Let's see.
```
public function verifyIban(Request $request)
    {
        return json_encode(self::doVerifyIban($request->input('iban'), $request->input('bic')));
    }
```

You might already see what is happening here that breaks everything, but let's take a look at doVerifyIban() now. I won't put the whole method, just the relevant parts. First we do this:
```
        if (! verify_iban($response->iban)) {
            $response->status = false;
            $response->message = 'Your IBAN is not valid.';

            return $response;
        }

        if (! iban_country_is_sepa(iban_get_country_part($response->iban))) {
            $response->status = false;
            $response->message = 'Your bank is not a member of SEPA (Single Euro Payments Area) so you can\'t use this bank account here. Please try another one.';

            return $response;
        }
```

Both of these do an excellent job of verifying the IBAN, whether it's a Dutch one or not. This is because the php-iban library is written quite well.

We have a bit of an issue here though, doVerifyIban, despite its name, also verifies BIC codes. You know, the one we weren't able to fill in yet? Oops!
```
            $country = substr($iban, 0, 2);

            if ($country == 'NL') {
                $response->bic = self::getNlBicFromIban($iban);
            }

            if (! self::verifyBicIsValid($response->bic)) {
                $response->status = false;
                $response->message = 'Your BIC is not valid.';

                return $response;
            }
```

Yeahh... If you're not from the Netherlands, the site can't automatically set a correct BIC for you, so it tries to verify the not-set BIC code, which obviously fails, and then returns a status of false. This is also why you get the message that your BIC is invalid when you enter your IBAN.

It should be noted that this function does do its job quite well other than this, it does actually verify BIC codes and IBANs correctly. You can check this by doing the **_allow internationals to sign up for Proto with their normal bank account glitch!_**.
Aka inspect element on the BIC field and remove the disabled from the element, lmao. There's also another funny way to glitch it out but I'll show that in the meeting.

TL;DR for the board: if you have an international that wants to sign up, inspect element on the BIC input field after entering someones IBAN and remove the "disabled" tag from the element. Now you can fill in the BIC code and even if it says it's invalid it will work.

MaxLiebe:
BIC field is now never disabled. BIC and IBAN are still validated in the back-end so there isn't really any potential for this to go terribly wrong.

Closes #2193 for now, verification should be improved in the controller itself.

### DIFF
--- a/resources/views/users/bankaccounts/addbank.blade.php
+++ b/resources/views/users/bankaccounts/addbank.blade.php
@@ -96,12 +96,12 @@
 
                             <div class="form-group">
                                 <label for="bic">Bank BIC Code</label>
-                                <input type="text" class="form-control text-uppercase" id="bic" name="bic" placeholder="" disabled>
+                                <input type="text" class="form-control text-uppercase" id="bic" name="bic" placeholder="">
                             </div>
 
-                            <p>
+                            <!-- <p>
                                 <span id="bic-message">Enter your IBAN first.</span>
-                            </p>
+                            </p> -->
 
                             <hr>
 
@@ -178,13 +178,13 @@
                     iban_message('black', 'We could not automatically verify your IBAN.')
                     bic_message('red', 'Please enter your BIC.')
                     bic.value = ''
-                    bic.disabled = false
+                    //bic.disabled = false
                 })
             } else {
                 iban_message('black', "Please enter your IBAN above.")
                 bic_message('black', 'Enter your IBAN first.')
                 bic.value = ''
-                bic.disabled = true
+                //bic.disabled = true
                 submit.disabled = true
             }
         })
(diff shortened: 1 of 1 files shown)


Jessseee (approved):
LGTM, I guess...

PR #2195, opened 2024-09-04
Update Proto.php to reflect the change to 14.0

ysbrandB:
Preparations for the GMM, should probably make these actual roles of a committee or put them in the .env but oh well.


Jessseee (approved):
LGTM, does 14.0 already have a picture for the birthday email?

PR #2199, opened 2024-09-06
Add participant limit column in activity overview

### LINKED ISSUE #2173: Include the sign up limit in the activity overview query
It would be amazing if the maximum amount of people that can sign up for an activity would be included in the Activity Overview query so we can see per activity what percentage of spot were filled:)

MaxLiebe:
Closes #2173


ysbrandB (approved):
LGTM

PR #2201, opened 2024-09-10
Move the google calendar ids from config to the .env

ysbrandB:
This makes them changeable without updating the config file. These have to be updated again with the new roster so would be nice to be able to update them easier.


Wilkuu-2:
There is no good way to actually test this on my side, but the code looks solid. 

LGTM?

PR #2215, opened 2024-09-19
Temporarily disable the feecron in October

ysbrandB:
This is due to the current system not being in sync with the new Rules and Regulations


Menkveld-24 (approved):
Lgtm

PR #2224, opened 2024-10-03
Fix the bankcontroller add method and the return type of getAdjacentPhoto

ysbrandB:
Small oversights in the big update.


Jessseee (approved):
LGTM

PR #2230, opened 2024-10-05
Add concurrency to the test and fix php action

ysbrandB:
This makes it so if you commit shortly after each other only the last one will run


ysbrandB:
This will be done with another PR

PR #2234, opened 2024-10-08
Add tests for dinnerform and search. Also fix some types

ysbrandB:
Fixed the return type of opensearch, and added a test for it.
Also fixes the types for the update email destination. 
Also adds tests for the adding of a dinnerform for members and the search pages and the needed factories for it.


Wilkuu-2 (approved):
LGTM

PR #2238, opened 2024-10-08
Fix crash on tickets::edit, when no event assigned

Wilkuu-2:
when no event was assigned to a ticket, the blade template crashed, because it did not check for it.

### DIFF
--- a/resources/views/tickets/edit.blade.php
+++ b/resources/views/tickets/edit.blade.php
@@ -34,8 +34,8 @@
                         <div class="form-group autocomplete">
                             <label for="event">Event:</label>
                             <input class="form-control event-search" id="event" name="event"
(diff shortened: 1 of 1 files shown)


ysbrandB:
Isnt that the desired behaviour though? All tickets should have an event associated with it.

Wilkuu-2:
There should be a way to enter the route, when that is the case, so that it can be fixed, though.

Wilkuu-2:
Also crashing the site by user interaction should never be desired behaviour.

ysbrandB:
A ticket can not be created without a ticket. This is never a supported use case. In the ticketcontroller we throw when there is no event associated with the ticket. With the update we only update it when there is a new event given. 
https://github.com/saproto/saproto/blob/0dc13182c3a74845e67cbae3f1ae9b3ad721641d/app/Http/Controllers/TicketController.php#L45

PR #2239, opened 2024-10-08
remove deprecated ChangeAndIfToEarlyReturnRector

Jessseee:


ysbrandB:
This is very outdated and has already been fixed in the #2213 update

PR #2242, opened 2024-10-10
Update to new Sephpa package conventions

MaxLiebe:
I'm guessing this package was updated somewhere during the Laravel 11 update. The correct syntax is used now and it should work again. I haven't been able to test it on a backup of the live website though.


ysbrandB (changes requested):
This does not handle the errors correctly. This addpayment method does not throw this exception anymore.

MaxLiebe:
LGTM, I can't review it because I made the PR 🥲

ysbrandB (approved):
LGTM

PR #2243, opened 2024-10-14
Disable the csp policy when on dev to correctly show the new laravel error page

ysbrandB:
Title says it all.


ysbrandB:
This way is now also recommended by the docs: https://github.com/spatie/laravel-csp?tab=readme-ov-file#using-whoops

PR #2244, opened 2024-10-15
Self host the free fontawesome files

ysbrandB:
Followed the guide at https://docs.fontawesome.com/web/setup/host-yourself/webfonts.
This adds the free and brand fontawesome icons from v6 with the compatability layers for v4 and v5.
This removes the dependancy on the free fontawesome kit which limits us to 10.000 pageviews per month.


Jessseee:
Why not use [@fortawesome/fontawesome-free](https://www.npmjs.com/package/@fortawesome/fontawesome-free), as @Wilkuu-2 did in https://github.com/saproto/saproto/pull/2182?

ysbrandB:
> Why not use [@fortawesome/fontawesome-free](https://www.npmjs.com/package/@fortawesome/fontawesome-free), as @Wilkuu-2 did in #2182? That would make it a lot easier to update in the future.

Now uses the package :)

PR #2251, opened 2024-10-18
Pre-fetch the font used on the site

ysbrandB:
Move the import of the font from css to the html head
This, combined with the prefetch makes it load in parallel with the css, instead of after it.


JonaMata (approved):
LGTM

PR #2252, opened 2024-10-21
Remove the old membership system

ysbrandB:
> [!caution]
> Only merge this pull-request after #2250 is merged and the commands ran and someone verified the database changes.

Removes the columns on the member table used by the old system and moves all business logic over to the new system.
The most important part it moves over is the feecron. Also removes the bridging code between the two systems.


Menkveld-24 (approved):
Jakub says the code looks nice

Menkveld-24:
Jk; actually looked through it and LGTM

Menkveld-24 (dismissed):
Lgtm

PR #2253, opened 2024-10-21
Fix the withdrawal XML

ysbrandB:
The sepa library does throw an exception when it says not to. See https://github.com/AbcAeffchen/Sephpa/issues/44.
Now we catch the potential error. 

Found some additional querybuilders we accidentally not call get on.
Added some extra BIC and IBAN validation when creating a bank using the SepaUtilities library.

Now the xml generation also performs some validation so if the IBAN and BIC do not match each other the treasurer will get notified when making the xml and it will have to be fixed before generating the xml.


Jessseee (approved):
LGTM

PR #2259, opened 2024-10-29
Change text to clarify events coming up (#2179)

### LINKED ISSUE #2179: The homepage says there are no events coming up (at all) while there are only no events coming up this month
image
See the image above, there are events, just not in the coming month. I think this is confusing and should be changed to either show the next n events or let the text reflect that there are no events in the coming month.

PringlePot:
Just a small change to make it more obvious what the "upcoming events" means. Fixes #2179


ysbrandB:
Did you check that the events we are giving the view actuall only contain the upcoming months activities? I think a nicer solution would then be to just show the upcoming 6 events and not do it on time.

Wilkuu-2 (approved):
LGTM

Issue #2091, opened 2024-02-01
NotUtwente Rule deprecation


Menkveld-24:
This validation rule appears to be deprecated

<img width="701" alt="afbeelding" src="https://github.com/saproto/saproto/assets/43108191/04187809-193b-471f-81c4-6906c4debe1f">

PringlePot:
Might be easier to check if the email just ends with "utwente.nl" or are there exceptions?

PR #2260, opened 2024-10-29
Rewrite utwente email check with new validation rules (#2091)

### LINKED ISSUE #2186: Issues with the signup of new users
- [ ] - the NotUtwente rule does not work, there are accounts being created with utwente emails.
- [ ] - /user/memberprofile/complete sometimes does not have the field phone every time, leading to: ErrorException: Undefined array key "phone" (does not get validated?).
- [ ] DEUTDE (deutsche bank) does not work with our sepa?
- [ ] When the dutch signup form for address is used it does not get checked if there is a valid response, so when it can not find it it tries to create a adress with null value. Leading to the following error: /user/address/add 1048 Column 'street' cannot be null.

PringlePot:
Currently just the same validator as before, but with the new ValidatorRule instead of depricated rule.

There is a big chance this will need some more chances to fix #2186

Also needs some tests to see if functionality stays the same (I can't test it without creating new accounts, which doesnt work yet lol)


ysbrandB (changes requested):
I think one check with !Str::endswith('utwente.nl') would be simpler and more maintainable. Then you can also ditch the second check.

PR #2277, opened 2024-11-16
Update composer and npm to update vulnerabilities

ysbrandB:
There are some Laravel vulnerabilities below the latest version. Therefore update the packages to the newest version. I believe the vuln does not work due to our PHP setup, but can not hurt to be sure anyways.


package.json
```diff
@@ -42,8 +42,8 @@
         "glob": "^7.1.6",
         "laravel-vite-plugin": "^0.7.8",
         "postcss": "^8.2.7",
-        "postcss-discard-comments": "^4.0.2",
```
Wilkuu-2, line 45: Nitpick: Major version update. Do those packages have breaking changes that affect us?

Wilkuu-2: I just checked, it's fine as far as I see.

Wilkuu-2 (approved):
LGTM

PR #2278, opened 2024-11-16
Fix the omnomcom wrapped endpoint

ysbrandB:
Fix the endpoints used by omnomcom wrapped. See also https://github.com/saproto/OmNomCom-Wrapped/pull/27.

### DIFF
--- a/routes/api.php
+++ b/routes/api.php
@@ -1,5 +1,6 @@
 <?php
 
+use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Route;
 
 Route::group(['middleware' => ['forcedomain'], 'as' => 'api::'], static function () {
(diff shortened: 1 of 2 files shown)


routes/api.php
```diff
@@ -13,7 +14,8 @@
     });
     /* Routes related to the User API */
     Route::group(['prefix' => 'user', 'as' => 'user::'], static function () {
-        Route::group(['middleware' => ['auth:api'], 'as' => 'qr::'], static function () {
+        Route::get('info', fn () => Auth::user())->name('info')->middleware('auth:api');
```
JonaMata, line 17: Why is this here?

JonaMata (changes requested):
LAGTM

routes/api.php
ysbrandB: This was first in the userapi controller, but that has been removed. It is used to check if the login was succesfull.

JonaMata (approved):
LGTM

PR #2279, opened 2024-11-17
Cleanup the unused flickr tables

ysbrandB:
The flickr tables are not used at all, and have all been migrated to the 'new' photo system. Therefore we do not need these anymore.


Wilkuu-2 (approved):
LGTM

PR #2284, opened 2024-11-23
Add cache control headers to .htaccess

ysbrandB:
When deploying our laravel folder gets overwritten. This also means the cache headers are taken out of the .htacces file. Therefore I think we should add it so it always gets deployed with our website. Already checked to be working by putting it into our .htaccess by hand between deploys. 

Rules taken from https://www.vimexx.nl/help/welke-headers-kan-ik-gebruiken-voor-optimalisatie-van-de-laadtijd#htaccess


ysbrandB:
This is already merged when merging vimexx to master

PR #2287, opened 2024-11-28
Sync aliases as groups

JonaMata:


app/Console/Commands/GoogleSync.php
```diff
@@ -271,6 +365,30 @@ public function listGoogleGroups(?ProtoUser $protoUser = null): Collection
         return $googleGroups;
     }
 
+    /**
+     * List Google Workspace groups. Pass in user to get only their groups.
```
Jessseee, line 369: Maybe update the description of this function.

app/Console/Commands/GoogleSync.php
```diff
@@ -158,6 +159,98 @@ public function syncGoogleGroupsWithCommittees(): void
...
+    {
+        $this->output->info('Alias Members:');
+
+        $aliases = ProtoAlias::query()->whereNotNull('destination')->get()->groupBy('alias');
+        $googleGroups = $this->listGoogleGroups();
+        $googleAliasGroups = $googleGroups->filter(fn ($group) => Str::startsWith($group->name, 'Alias'));
+
+        foreach ($aliases as $aliasGroup) {
+            $this->pp(
+                '<fg=green>✓</> '.str_pad("$aliasGroup->first()->alias <fg=gray>", 65, '.')."</> $aliasGroup->count() members",
```
ysbrandB, line 223: This print does not work, you can only acces the variable in this way and not the children.

Jessseee (approved):
LGTM

PR #2308, opened 2024-12-11
get x of achievment fixed

tijmi:
fixed that for achievements where you need to buy x of that you don;t ahve to get 1 more to get the achievement


app/Console/Commands/AchievementsCron.php
```diff
@@ -13,7 +13,8 @@
 use App\Models\Product;
 use App\Models\ProductCategory;
 use App\Models\User;
-use Carbon;
+// use Carbon;
```
ysbrandB, line 16: Please remove this comment, or properly update the 'use'

ysbrandB (changes requested):

Issue #2301, opened 2024-12-03
Change the omnomcom total to only show the total of products that can be bought in the omnomcom


ysbrandB:
The current total amount of money spent shown in the omnomcom when you pay now includes activities. This is weird behaviour if the treasurer just processed an activity, you go to buy a snicker and it says you have spent 176,28 euros today. That seems like an expensive snicker. I think it should be changed to only add the sum of products whose category are in the omnomcom or tipcie stores.

PR #2318, opened 2025-01-08
Limit omnomcom "Total spent" to only bought items (#2301)

PringlePot:
Restrict the total spent calculation to include only items that belong to the purchased categories. Fixes issue #2301


PringlePot:
Would be nice if someone checked this since im not 100% sure of the SQL query

app/Http/Controllers/OmNomController.php
```diff
@@ -251,7 +251,20 @@ public function buy(Request $request, string $store_slug)
...
 
             if ($user->show_omnomcom_total) {
-                $result->message = sprintf('You have spent a total of <strong>€%0.2f</strong>', OrderLine::query()->where('user_id', $user->id)->where('created_at', 'LIKE', sprintf('%s %%', date('Y-m-d')))->sum('total_price'));
+                $categories = collect($stores)->pluck('categories')->flatten()->unique();
+
+                $totalSpent = OrderLine::query()
+                    ->where('user_id', $user->id)
+                    ->where('created_at', 'LIKE', sprintf('%s %%', date('Y-m-d')))
+                    ->whereHas('product.categories', function ($query) use ($categories) {
+                        $query->whereIn('categories.id', $categories);
```
ysbrandB, line 260: categories.id fails, because categories is not a table in the db. It should probably be something like product_categories, but look at the model what it should be.

PringlePot: yup its called `product_categories` updating it.

ysbrandB (approved):
LGTM

PR #2321, opened 2025-01-13
Fix unnecessary ProTube call when removing role and add error handling

### LINKED ISSUE #2200: Removing a role sends unnecesary checks to protube without catch logic
When an authorization role is removed (even when it is not the protube role) a call is made to protube. The logic for how to handle a failure behind the protube call logic is also missing. When protube is down it results in the following error: 
image

PringlePot:
Removing a role now only triggers a ProTube API call when the role is specifically the ProTube role. Additionally, error handling has been implemented to manage potential failures when the ProTube service is unavailable.

Fixes #2200

### DIFF
--- a/app/Http/Controllers/AuthorizationController.php
+++ b/app/Http/Controllers/AuthorizationController.php
@@ -3,6 +3,7 @@
 namespace App\Http\Controllers;
 
 use App\Models\User;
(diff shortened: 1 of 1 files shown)


ysbrandB (approved):
LGTM

PR #2334, opened 2025-01-28
Disable lazyloading of attributes when in dev

ysbrandB:
Add the line `Model::preventLazyLoading(! app()->isProduction());` to our AppserviceProvider.
This makes it so when in dev the framework will throw an error if an attribute is lazyloaded.
Also already fixed a bunch of the issues that break pages because of this.

### DIFF
--- a/app/Console/Commands/AchievementsCron.php
+++ b/app/Console/Commands/AchievementsCron.php
@@ -336,7 +336,7 @@ class AchievementsCron extends Command
     {
         $products = [];
         foreach ($categories as $category) {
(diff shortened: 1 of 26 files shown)


ysbrandB:
> ## Suspect Issues
> This pull request was deployed and Sentry observed the following issues:
> 
> * ‼️ **BadMethodCallException: Method Illuminate\Database\Eloquent\Collection::exists does not exist.** `/user/{id}/member/impersonate` [View Issue](https://sa-proto.sentry.io/issues/6261189927/?referrer=github-pr-bot)
> 
> Did you find this useful? React with a 👍 or 👎

Also the past events do not show up anymore

PR #2337, opened 2025-02-02
Replace the smartxp screen with the protopolis screen

ysbrandB:
Replace the smartxp screen with the protopolis screen, but instead of the narrowcasting iframe it has an i-frame of the protube screen. 

Now looks like this:
image


app/Http/Controllers/SmartXpScreenController.php
```diff
@@ -114,7 +98,9 @@ public function smartxpTimetable()
     /** @return View */
     public function canWork()
     {
-        return view('smartxp.caniwork', ['timetable' => $this->smartxpTimetable()->roster,
-            'occupied' => $this->smartxpTimetable()->occupied, ]);
+        $timetable = $this->smartxpTimetable();
+
+        return view('smartxp.caniwork', ['timetable' => $timetable->roster,
+            'occupied' => $timetable->occupied, ]);
```
MaxLiebe, line 104: Does this even work right now now that the university has switched to TimeEdit?

MaxLiebe (approved):
LGTM

app/Http/Controllers/SmartXpScreenController.php
ysbrandB: Yes, we sync timedit via the ical link to our own google calendar

PR #2346, opened 2025-03-03
WIP: Update to Laravel 12 ('Vo!)

ysbrandB:
This updates all the necessary packages to Laravel versie 12. (https://laravel.com/docs/12.x/upgrade)

The most impactfull change is the update from Carbon 2 to 3. (https://carbon.nesbot.com/docs/#api-carbon-3)
- The createFromTimestamp method now uses UTC by default, and as we use Europe/Amsterdam we now need to pass in that timezone everywhere we use the function. 
- The diffIn methods return a signed float now instead of an integer, which means we have to cast it to an int and set absolute to true to retain the old behaviour.

Further Changes:
- Updated Larastan (https://github.com/larastan/larastan/blob/3.x/UPGRADE.md)
- Updated Rector, and add new Rulesets
- Removed Laravel Solo as for now it is not working in docker
- Update graham-campbell/markdown (no changes)
- Update milon/barcode (no changes)
- Update vinkla/hashids (no changes)


ysbrandB:
This is not yet finished, Carbon also changed the default of Create::('invalid') from false to null, so a bunch of checks will now fail by default. Like from our CommitteeController: if (($membership->created_at = Carbon::create($request->start)) === false) {} now always returns true

ysbrandB:
This work has been done in smaller, seperate PR's, and is closed by: https://github.com/saproto/saproto/pull/2401

PR #2362, opened 2025-03-22
Feature: Proto Sticker Tracker

ysbrandB:
Have you ever encountered a Proto sticker in a cool place? 
Wouldnt it be cool to have a map of all the places we have already conquered?
This PR adds a map on which members can place a pin where they have put a Proto sticker, and then get a popup to include a photo of the sticker. Then you can see the sticker as a pin on the map and when you click on it the photo of the sticker (see images).

This PR also adds achievements for your first sticker placed, and when you have logged stickers in 3, 6 and 12 distinct countries respectively. 

image

image


PringlePot:
Would it be fun to also add support for do group stickers? I know a lot of do group still use them even way after the kick in.

Issue #2373, opened 2025-04-02
Photo admin sorts from oldest to newest published album


badrboubric:
Title says it all (Again)

![Image](https://github.com/user-attachments/assets/5b75384f-1f0d-4912-90fd-d4fe1c4f0539)

PR #2379, opened 2025-04-08
Proper sorting for committee page

### LINKED ISSUE #2325: Committee's "Previously Organized" and "Events Previously helped at" show old events
image

Since these events are (mostly) not relevant anymore, it might be a good idea to only limit to events this year? Another possibility could to have a "view more" button and page where all of the events that the committee has organized are displayed

PringlePot:
Fixes #2325, #2373

Gotta love resort()


ysbrandB (approved):
LGTM

PR #2380, opened 2025-04-23
Fix registrations made within a second landing on back-up list.

Wilkuu-2:
This single line change fixes the problem that anyone that happens to be fast enough to register within a second of the opening ends up on the back-up list of an event.

I caught due to some creative end-to-end testing in production.

```
   0|           1s          |              ... 
    ^       ^               ^ 
  opens   back-up         signed up
 ```

### DIFF
--- a/app/Models/Activity.php
+++ b/app/Models/Activity.php
@@ -249,7 +249,7 @@ class Activity extends Validatable
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


Menkveld-24 (approved):
Lgtm

ysbrandB (approved):
LGTM

PR #2388, opened 2025-05-05
added new font to omnomcom

tijmi:


resources/views/website/assets/stylesheets.blade.php
```diff
@@ -12,7 +12,11 @@
 {{-- load the font used on our website and prefetch it --}}
 <link rel="preconnect" href="https://fonts.googleapis.com" />
 <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+<link rel="preconnect" href="https://fonts.googleapis.com" />
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+<link rel="preconnect" href="https://fonts.googleapis.com" />
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
 <link
-    href="https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,100;0,300;0,400;0,700;0,900;1,100;1,300;1,400;1,700;1,900&display=swap"
+    href="https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,100;0,300;0,400;0,700;0,900;1,100;1,300;1,400;1,700;1,900&family=Varela+Round&display=swap"
```
ysbrandB, line 20: Do we need to add it here? This will load the font on all pages, not just the omnomcom ones

resources/views/omnomcom/minisite.blade.php
```diff
@@ -6,7 +6,7 @@
 <html lang="en">
     <head>
         <link
-            href="https://fonts.googleapis.com/css?family=Roboto+Slab:400,700"
+            href="https://fonts.googleapis.com/css2?family=Varela+Round&display=swap"
```
ysbrandB, line 9: I think the inclusion here should be enough, then we do not need it in stylesheets.blade.php right?

ysbrandB (approved):
LGTM

PR #2389, opened 2025-05-07
Update: Password validation and requirements

ysbrandB:
This PR moves our password requirements from:
-  longer than 10 characters

To:
- longer than 10 letters
- Must include upper and lower case letters
- Must include a number
- Must include a symbol
- May not be compromised in a data leak

For the may not be compromised in a data leak it now uses the built in laravel validator, instead of the seperate library.
It also adds back in the syncing of the password to google, which was accidentally removed in the SAML update.


PringlePot:
I looked into it. Bcrypt (laravels default hashing algorithm) truncates input to the first 72 characters. We should probably add a limit for 72 characters for this reason, since longer passwords will only be checked for the first 72 characters.

app/Providers/AppServiceProvider.php
```diff
@@ -20,6 +21,14 @@ public function boot(): void
     {
         Paginator::useBootstrapFive();
 
+        Password::defaults(function () {
+            $rule = Password::min(10)->letters();
```
PringlePot, line 25: a max could be added here too with the max() if im not mistaking

ysbrandB:
> I looked into it. Bcrypt (laravels default hashing algorithm) truncates input to the first 72 characters. We should probably add a limit for 72 characters for this reason, since longer passwords will only be checked for the first 72 characters.

If it automatically truncates it anyway I do not really see a point in limiting it. It does not try to encrypt more than those 72 characters anyways right?

PringlePot:
> > I looked into it. Bcrypt (laravels default hashing algorithm) truncates input to the first 72 characters. We should probably add a limit for 72 characters for this reason, since longer passwords will only be checked for the first 72 characters.
> 
> If it automatically truncates it anyway I do not really see a point in limiting it. It does not try to encrypt more than those 72 characters anyways right?

this means that if someone only changes the last characters in a longer password, the old one will keep working. Shouldn't be a problem in most cases but might cause some unexpected behaviour

ysbrandB:
> > > I looked into it. Bcrypt (laravels default hashing algorithm) truncates input to the first 72 characters. We should probably add a limit for 72 characters for this reason, since longer passwords will only be checked for the first 72 characters.
> > 
> > 
> > If it automatically truncates it anyway I do not really see a point in limiting it. It does not try to encrypt more than those 72 characters anyways right?
> 
> this means that if someone only changes the last characters in a longer password, the old one will keep working. Shouldn't be a problem in most cases but might cause some unexpected behaviour



> > > I looked into it. Bcrypt (laravels default hashing algorithm) truncates input to the first 72 characters. We should probably add a limit for 72 characters for this reason, since longer passwords will only be checked for the first 72 characters.
> > 
> > 
> > If it automatically truncates it anyway I do not really see a point in limiting it. It does not try to encrypt more than those 72 characters anyways right?
> 
> this means that if someone only changes the last characters in a longer password, the old one will keep working. Shouldn't be a problem in most cases but might cause some unexpected behaviour

Yeah fair, added now

PringlePot (approved):

app/Http/Controllers/ActivityController.php
```diff
@@ -169,10 +165,9 @@ public function addHelp(Request $request, Event $event)
     }
 
     /**
-     * @param  int  $id
      * @return RedirectResponse
      */
-    public function updateHelp(Request $request, $id)
+    public function updateHelp(Request $request, int $id)
```
Menkveld-24: Je zou hier implicit model binding kunnen toevoegen, maar dit werkt ook

app/Http/Controllers/AuthController.php
```diff
@@ -141,15 +123,10 @@ public static function verifyCredentials(string $username, string $password)
 
         if ($user == null) {
             $member = Member::query()->where('proto_username', $username)->first();
-            $user = ($member ? $member->user : null);
+            $user = ($member?->user);
```
Menkveld-24:
```suggestion
            $user = $member?->user;
```

app/Http/Controllers/UserController.php
```diff
@@ -26,26 +26,10 @@ public function destroy(Request $request): RedirectResponse
...
         if (Auth::id() == $user->id) {
-            $password = $request->input('password');
-            $auth_check = AuthController::verifyCredentials($user->email, $password);
-            if ($auth_check == null || $auth_check->id != $user->id) {
-                Session::flash('flash_message', 'You need to provide a valid password to deactivate your account.');
-
-                return Redirect::back();
-            }
+            $request->validate(
+                ['password' => ['required', 'current_password']]
```
Menkveld-24: Wist niet dat dit een ding was, nice!

config/hashing.php
```diff
@@ -29,7 +29,7 @@
     */
 
     'bcrypt' => [
-        'rounds' => env('BCRYPT_ROUNDS', 10),
+        'rounds' => env('BCRYPT_ROUNDS', 12),
```
Menkveld-24: Kan je dit zomaar aanpassen zonder oude wachtwoorden te slopen?

ysbrandB: Yes, wordt gerehashd op login

PR #2391, opened 2025-05-08
Fix not being able to force sign up a helper if the user is signed in to the event itself

ysbrandB:
This PR cleans up the participation controller. 

It remove the is_web check (we do not support the app anymore, so all our traffick comes through the Auth::web guard).
It also fixes that you are not able to sign up a helper for an event if you are in the event itself.
Additionally, it moves the controller over to route_model binding.


Wilkuu-2 (approved):
LGTM

PR #2480, opened 2025-09-24
Move omnomcomsounds to media library

JonaMata:
> [!CAUTION]
> Run `php artisan proto:migrate-om-nom-com-sounds` to migrate the existing omnomcom sounds to the new storage location.


ysbrandB (approved):
LGTM

PR #2497, opened 2025-10-21
Styling/consistent route syntax

ysbrandB:
Moves over our web.php file from the Route::controller syntax to [Controller::class, 'function']. This is better supported in editors and makes more clear which functions get used or not. 

With the new syntax I have already discovered two routes that should not exist anymore: 
Route::post('edit/{id}/image', [NewsController::class,'featuredImage'])->name('image') and Route::get('edit/{id}', [FeedbackController::class,'categoryEdit'])->name('edit'); These are now removed

### DIFF
--- a/app/Http/Controllers/EmailController.php
+++ b/app/Http/Controllers/EmailController.php
@@ -199,7 +199,7 @@ class EmailController extends Controller
 
             $email->ready = true;
             $email->save();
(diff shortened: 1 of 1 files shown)


Wilkuu-2:
This is adding so much boilerplate, but if it improves people's UX it's somewhat justified. My IDE seems to detect the pre-PR route tree just fine.

PR #2520, opened 2025-11-20
Switch everything over to yarn

### LINKED ISSUE #2519: Change: Move to Yarn instead of npm
Yarn can work in parallel, so it would improve the build speed. The only requirement would be to have an up-to-date yarn lock file.

PringlePot:
Fixes #2519

Needs some proper testing, since I am not 100% sure about the github workflows. Also added vite to the dependencies since it was not already.


ysbrandB (changes requested):
The install still gives me the message: warning " > laravel-echo@2.2.6" has unmet peer dependency "socket.io-client@*".

PR #2542, opened 2026-01-05
Remove the softdeletes from the activities_users table and add a unique constraint

ysbrandB:
Adds a unique constraint to the activities_users table so we know for sure users can not sign up twice.


ysbrandB:
This will be done after https://github.com/saproto/saproto/pull/2597

PR #2692, opened 2026-05-18
Send a reminder email to the treasurer when there are unclosed activities

### LINKED ISSUE #2636: Reminder email to close an activity for the treasurer
Now sometimes the treasurer forgets to close activities in the website.
It would be nice if they got an email when an activity has happened in the past that they need to close it, with the link to the page where to close the activities.
This could for instance be one month after an activity has happened, and it has not been closed yet

ysbrandB:
This PR sends an email to the treasurer when there are unclosed activities that ended more than a month ago.
It sends a list of those activities as a reminder to close them, so this does not get forgotten. It will send an email every monday at 8.
Fixes #2636.


Menkveld-24 (dismissed):
Lgtm

